Issue #57, opened 2016-11-28
Line Length


allanj-uaag:
<h2>SC Shortname</h2>
<p>Line Length </p>
<h2>SC Text</h2>
<p>For the visual presentation of all text, a mechanism is available such that line length is user adjustable, to 25 characters, with no two-dimensional scrolling required, and with the following exceptions.</p>
<ol>
  <li> If the user-agent fits the layout to the view port, and does not provide a means of re-flowing content, then two-dimensional scrolling is exempt. </li>
  <li> If the spatial layout of text is essential to its use, then two-dimensional scrolling is exempt. </li>
</ol>
<h2>Suggested Priority Level</h2>
<p>Level A </p>
<h2>Related Glossary additions or changes</h2>
<p>none</p>
<h2>What Principle and Guideline the SC falls within.</h2>
Principle 1, Guideline <a href="http://www.w3.org/TR/WCAG20#visual-audio-contrast-visual-presentation 1.4.8 Visual Presentation"> 1.4.8 Visual Presentation</a>
<h2>Description</h2>
<p>The user can pick the length of lines to fit reading needs. Many user agents already support this using reflow and adjustable viewport size. </p>
<img src="https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/images/7/7e/LinLenHTML.png" alt="A reflowed HTML file with reduced line length."/>
<p> This shows how a line length can be shortened by reducing the browser width. The page has been reflowed using a custom style sheet made available through a browser extension.</p>
<p><img src="https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/images/thumb/6/65/LinLenPDF.png/300px-LinLenPDF.png" alt="A reflowed display with reduced line length in PDF."/></p>
<p>The same works with reflowed PDF.</p>
<p>Character count was chosen as a metric to measure line length because a character count enables line-length control for both large print with normal width, and narrow columns with normal print. The number 25 was chosen to fall in a range above the length of larger words in most languages. (<a href="http://www.ravi.io/language-word-lengths">Distribution of word lengths in various languages</a>.) This reduces excessive hyphenation. On a 13-inch monitor, one can use 33 point font and fit in 25 characters. Character count size is 72(13cos(30))/25.</p>
<h2>Benefits</h2>
[...]

PR #85, opened 2017-01-11
Issue 57 Line Length SC

DavidMacDonald:
I left it as 25 characters line length, although some thought that was
too short a line to require (45-65 characters suggested). I also added
“Blocks of text” as a compromise with James Nurthan’s concern. See
https://github.com/w3c/wcag21/issues/57.

This is a test.


awkawk:
David, I don't think that worked. I'm not sure what went wrong but it is showing the guidelines index as initially empty and then your changes bring in all of the content.  Odd.

Also, you should be making the pull request on the FPWD branch.

DavidMacDonald:
I see what I did wrong... I cloned it to my local Github desktop from here rather than forking it to my Github repro and cloning it from there.

DavidMacDonald:
You can delete it.

PR #87, opened 2017-01-11
Issue 57 Line Length

DavidMacDonald:
I kept 25 characters although there are some who feel it should be
45-60 characters.  changed it to blocks of text on James comment about
the difficultly in applying it to ALL text.


DavidMacDonald:
Delete this... will pull against FPWD

Issue #58, opened 2016-11-28
Reflow to Single Column


allanj-uaag:
# Current versions of SC and Definitions
- [SC for viewing](https://rawgit.com/w3c/wcag21/linearization_ISSUE-58/guidelines/sc/21/linearization.html) | [SC for editing](https://github.com/w3c/wcag21/blob/linearization_ISSUE-58/guidelines/sc/21/linearization.html)
- [SC in full draft guideline](https://rawgit.com/w3c/wcag21/linearization_ISSUE-58/guidelines/#linearization) ## Current versions of SC and Definitions
- [Understanding doc for viewing](https://rawgit.com/w3c/wcag21/linearization_ISSUE-58/understanding/21/linearization.html) | [Understanding doc for editing](https://github.com/w3c/wcag21/blob/linearization_ISSUE-58/understanding/21/linearization.html) | [Understanding doc in master](https://rawgit.com/w3c/wcag21/master/understanding/21/linearization.html)
## Open issues and Surveys
Open issues: [SC Status page](https://www.w3.org/WAI/GL/wiki/WCAG_2.1_SC_status)

- 
(Links to surveys require W3C Member access)

<h2>SC Shortname</h2>
<p>Linearization </p>
<h2>SC Text</h2>
<p>A mechanism is available to view content as a single column, except for parts of the content where the spatial layout is essential to the function and understanding of the content.</p>
<h2>Suggested Priority Level</h2>
<p>Level A </p>
<h2>Related Glossary additions or changes</h2>
<p>none</p>
<h2>Relevant Guidelines and Success Criteria</h2>
<ul>
  <li><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation.html"> Guideline 1.3 Flexible Data</a>
    <ul>
      <li><a href="https://www.w3.org/TR/WCAG20/#content-structure-separation-sequence"> SC 1.3.2 Meaningful Sequence </a></li>
    </ul>
  </li>
  <li><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast.html"> Guideline 1.4 Distinguishable </a></li>
  <ul>
    <li><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html"> SC 1.4.4 Resize Text </a></li>
[...]

PR #88, opened 2017-01-12
Issue57 line length

DavidMacDonald:
This is where we are so far, seems there is a fair amount of discussion continuing.


DavidMacDonald:
This can be closed, the LVTF has retired issue #57 and will address the use case with #58

PR #91, opened 2017-01-23
Add initial Accessible Authentication SC

JohnRochfordUMMS:
https://github.com/w3c/wcag21/issues/23

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -689,6 +689,14 @@
 					<p class="change">Unchanged</p>
 					<p><a>Section</a> headings are used to organize the content.</p>
 				</section>
+        
+        <section class="sc">
+	        <h4>Accessible Authentication</h4>
+	        <p class="conformance-level">A</p>
+	        <p class="change">Unchanged</p>
+	        <p>A user authentication method is offered that does not rely upon a user's ability to memorize information; recall information from memory; speak; or mentally process presented or recalled information beyond the mental processes that are required to use a simple web page.</p>
+         </section>
+        
 			</section>
 		</section>
 


JohnRochfordUMMS:
https://github.com/w3c/wcag21/issues/23

awkawk:
Lisa mentioned: "https://www.w3.org/TR/webauthn/ and https://www.w3.org/Webauthn/is the web authetification that is refered to in the techneques . Conforming to it enables complete conformance"

This spec is for user agents to conform to though, not authors of content, so I'm not sure that helps.

patrickhlauke:
This SC seems to confuse authentication (in the sense of being able to log into one's account through a unique username/password, for instance) with CAPTCHAs

> Many user authentication methods rely upon trying to differentiate between a human, and software (bots) that try to pose as a human. 

This sounds a lot like the problem this tries to tackle is indeed CAPTCHAs. And these are quite distinct from "authentication".

As an initial step, I'd like to see this clearly disambiguated in the SC text.

patrickhlauke:
In most cases ( ? ) CAPTCHAs already fail various current 2.0 SCs (e.g. visual CAPTCHAs fail 1.1.1). If indeed we're talking CAPTCHAs, then it would be good to make a gap analysis of where certain types of CAPTCHA are currently not adequately covered by 2.0 (for instance, those based on quizzes/mathematical equations).

patrickhlauke:
Specs like https://www.w3.org/TR/webauthn/ are designed to deal with the "uniquely identifying a user" problem of traditional login (username/password) systems. As such, they solve a potentially different problem (e.g. I may not want to divulge my uniquely identifiable information to a site that only wants to know if I'm a human or not). Unless I'm missing some basic functionality in that spec that can be used as a very basic "the user is who they say they are" which can be interpreted as "the user is a real user".

patrickhlauke:
(incidentally, is this the right place to comment, or the issue itself?)

jnurthen:
I will also note that this specification is currently in draft state. While
I understand that it may reach Recommendation before WCAG 2.1 - until it
does so we have no idea how widely it will be adopted by browsers. I am
uncomfortable relying on a draft spec like this in order to enable
conformance.

lseeman:
WCAG 2.0 came out when ARIA was just a WD. It is good enough
Also it is how you solve unique identification. There are other techniques for capture

patrickhlauke:
If you ARE trying to address both unique identification (logging into one's own account) and challenges like CAPTCHA, then i'd suggest these should be split out into two separate SCs, as they are fundamentally different:

- unique identification is about providing something the unique user knows (e.g. a password, which images to select in a set of images, which pattern to draw, etc) and/or something the user has (a specialised security fob/key, their fingerprint/similar biometrics)
- CAPTCHAs are about proving the user is not a robot by performing a task which is currently not possible to unambiguously achieve computationally

techniques to satisfy these may be similar, but the problem itself is different enough to warrant separation, i'd say

lseeman:
yes, we could seprate it

patrickhlauke:
worth adding that one aspect in which the two differ is in their security considerations: if a CAPTCHA is insecure/easily fooled, then you'd get spam/fake entries in an opinion poll or similar; however, if a unique identification can be too easily circumvented, you'd get potentially somebody gaining access to a user's financial information, electronic voting fraud, etc.

jasonjgw:
If the user agent doesn't support biometric authentication techniques, hardware authentication tokens (which are generally offered as part of multi-factor schemes rather than as the only authentication method, i.e., complementing but not supplanting user names/passwords), or another alternative, then it's far from clear what the content author can do about it while maintaining security. Failing to maintain security isn't an acceptable option: it would leave users' personal information, finances, etc., all vulnerable to compromise. If someone runs into accessibility issues with user names and passwords for cognitive reasons, then I worry that they may well be among the very group of people who could least effectively deal with the consequences of having their personal information, access to their finances, etc., exploited by bad actors.

Password managers in user agents are widely implemented, but it is still necessary to create and store a password initially for each authenticated Web site or application. Thus, I think password managers limit the problem without solving it entirely.

I think this is really a user agent requirement, but one that content authors can support when the appropriate APIs become available. I agree that reliably distinguishing human users from programs is a separate problem (giving rise to CAPTCHA) and that it should be treated as independent of the authentication issue.

The tendency in practice, for security reasons, is to move toward two-factor authentication ("what you have" as well as "what you know"), which can increase the cognitive demands to some extent - but anyone who operates a secure application has good reasons to move beyond merely relying on passwords.

Has anyone performed a security analysis of proposed means of implementing this proposal?

awkawk:
Old merge, overcome by events.

PR #98, opened 2017-01-30
added interuptions

lseeman:
and the glossary for easily availible

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -689,6 +689,13 @@
 					<p class="change">Unchanged</p>
 					<p><a>Section</a> headings are used to organize the content.</p>
 				</section>
+        
+        <section class="sc">
+					<h4>Interruptions (minimum)</h4>
+					<p class="conformance-level">AA</p>
+				<p>There is an easily available mechanism to postpone and suppress interruptions and changes in content unless they are initiated by the user or involve anֲ emergency.</p>
+				</section>
+        
 			</section>
 		</section>
 
@@ -1170,6 +1177,15 @@
 					<p>audio description that is added to an audiovisual presentation by pausing the <a>video</a> so that there is time to add additional description</p>
 					<p>This technique is only used when the sense of the <a>video</a> would be lost without the additional <a>audio description</a> and the pauses between dialogue/narration are too short.</p>
 				</dd>
+        <dl>
+		          <dt><dfn>easily available (or easily available mode or setting) one or more of the following are true:</dfn></dt>
+                  <dd>
+        <ul>
+  <li>can be set one time with as wide a scope as possible (such as using the standards of the OS, From ISO 9241-112 or GPII when available) </li>
+  <li>has the option to save or change the setting, where available interoperably, but also for the scope of the set of web pages </li>
+  <li>is reachable from each screen where it may be needed, and the path and the control conforms to all of this document</li>
+  </ul></dd></dl>
+        
 				<dt><dfn>flash</dfn></dt>
 				<dd>
 					<p>a pair of opposing changes in <a>relative luminance</a> that can cause seizures in some people if it is large enough and in the right frequency range</p>


awkawk:
@lseeman can you confirm that this is related to Issue 47?

lseeman:
yes. (as much as i can follow this)

moekraft:
- Editorial:  There is an extra character between "an" and "example", i.e. or "involve anֲ emergency."

awkawk:
Pull request on https://github.com/w3c/wcag21/issues/47 topic.

guidelines/index.html
```diff
@@ -1170,6 +1177,15 @@
 					<p>audio description that is added to an audiovisual presentation by pausing the <a>video</a> so that there is time to add additional description</p>
 					<p>This technique is only used when the sense of the <a>video</a> would be lost without the additional <a>audio description</a> and the pauses between dialogue/narration are too short.</p>
 				</dd>
+        <dl>
+		          <dt><dfn>easily available (or easily available mode or setting) one or more of the following are true:</dfn></dt>
+                  <dd>
+        <ul>
+  <li>can be set one time with as wide a scope as possible (such as using the standards of the OS, From ISO 9241-112 or GPII when available) </li>
+  <li>has the option to save or change the setting, where available interoperably, but also for the scope of the set of web pages </li>
+  <li>is reachable from each screen where it may be needed, and the path and the control conforms to all of this document</li>
```
steverep: * Replace "screen" with "webpage", which has an existing glossary definition
* Delete "and the path and control conforms to all of this document".  This is not only implied in my opinion, but also does not specify a conformance level and conflicts/overlaps with Note 2 in the [definition of mechanism](https://www.w3.org/TR/WCAG20/#mechanismdef).
* "Reachable" is not defined and so this requirement is very vague.  Technically, if I need to click Settings and then jump through 10 pages of controls to get to the mechanism for suppressing the interruption, that is reachable.

If the intent here is to be able to suppress with a single click, then maybe something like: "able to be set without any change of content".

guidelines/index.html
```diff
@@ -1170,6 +1177,15 @@
 					<p>audio description that is added to an audiovisual presentation by pausing the <a>video</a> so that there is time to add additional description</p>
 					<p>This technique is only used when the sense of the <a>video</a> would be lost without the additional <a>audio description</a> and the pauses between dialogue/narration are too short.</p>
 				</dd>
+        <dl>
+		          <dt><dfn>easily available (or easily available mode or setting) one or more of the following are true:</dfn></dt>
+                  <dd>
+        <ul>
+  <li>can be set one time with as wide a scope as possible (such as using the standards of the OS, From ISO 9241-112 or GPII when available) </li>
+  <li>has the option to save or change the setting, where available interoperably, but also for the scope of the set of web pages </li>
```
steverep: The latter part of this requirement is dangling (i.e. "but also..." only contains a prepositional phrase).  I'd like to help with rewording, but I'm confused by the intent of this bullet versus the previous one.  They both seem to be dealing with scope of the setting.  Can you clarify the distinction?

guidelines/index.html
```diff
@@ -1170,6 +1177,15 @@
 					<p>audio description that is added to an audiovisual presentation by pausing the <a>video</a> so that there is time to add additional description</p>
 					<p>This technique is only used when the sense of the <a>video</a> would be lost without the additional <a>audio description</a> and the pauses between dialogue/narration are too short.</p>
 				</dd>
+        <dl>
+		          <dt><dfn>easily available (or easily available mode or setting) one or more of the following are true:</dfn></dt>
+                  <dd>
+        <ul>
+  <li>can be set one time with as wide a scope as possible (such as using the standards of the OS, From ISO 9241-112 or GPII when available) </li>
```
steverep: In the spirit of accessibility, expand the acronyms for OS and GPII.

michael-n-cooper:
Addressed by 2705d25a36fe8fd7b5025d6cf72133843c55a99c. Got some of the editorial comments but not the more substantial ones. Further process of #47 should come back to this.

Issue #51, opened 2016-11-27
Visual presentation


lseeman:
# SC Shortname: Visual Presentation
## SC Text

Allow for the ability to easily adjust white space around objects and text, including boxes, paragraph headings, and content, to a degree that suits the user and does not disrupt the overall integrity of a web page.

# Current:

1.4.8 Visual Presentation: For the visual presentation of blocks of text, a mechanism is available to achieve the following (Level AAA):

1.    Foreground and background colors can be selected by the user.

2.    Width is no more than 80 characters or glyphs (40 if CJK).

3.    Text is not justified (aligned to both the left and the right margins).

4.    Line spacing (leading) is at least space-and-a-half within paragraphs; and paragraph spacing is at least 1.5 times larger than the line spacing.

5.    Text can be resized, without assistive technology, up to 200 percent in a way that does not require a user to scroll horizontally to read a line of text on a full-screen window.

 

# Proposed:

1.4.8 Visual Presentation: For the visual presentation of blocks of text and objects, a mechanism is available to achieve the following (Level AA):

1.    Foreground and background colors can be selected by the user.

2.    Width is no more than 80 characters or glyphs for Latin and Semitic based languages; or 40 for Chinese, Japanese, and Korean; or can be selected by the user.

3.    Text is not fully justified (aligned to both the left and the right margins), or justification can be set by the user.

4.    Line spacing (leading) is at least space-and-a-half within paragraphs; and paragraph spacing is at least 1.5 times larger than the line spacing.

5.    Text can be resized, without assistive technology, up to 200 percent in a way that does not require the user to scroll horizontally to read a line of text on a full-screen window.

6.    Increased line and border spacing can be added around blocks of text and objects, such that they can be increased up to 200% without loss of content or functionality.

Include https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-visual-presentation.html

 

# Suggestion for Priority Level (AA)

AA

As this can be enabled via personalization, it is widely applicable.

##Related Glossary additions or changes

 
##What Principle and Guideline the SC falls within.

Guideline 1.4 Distinguishable: Make it easier for users to see and hear content, including separating foreground from background.

Or under the personalization guideline

# Description

White space, also described as negative space, reduces clutter and provides definition to content so that the viewer is given a clear overview of a web page. It is used by designers to enhance text and the position of objects on a page.

# Benefits

[...]

PR #113, opened 2017-02-09
Spacing - issue 51

lseeman:
Issue is at
https://github.com/w3c/wcag21/issues/51

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -489,6 +489,13 @@
 					<p><a>Images of text</a> are only used for <a>pure decoration</a> or where a particular presentation of <a>text</a> is <a>essential</a> to the information being conveyed.</p>
 				</section>
 			</section>
+      
+      <section class="sc">
+							<h4>Spacing</h4>
+						<p class="conformance-level">AA</p>
+							<p>Line spacing between paragraphs, blocks of text, graphics and embedded elements and objects and regions is at least 0.8em vertically and 0.5em horizontally and line spacing in cursive script languages have a line high of at least 150% unless a mechanism is available such that they can be increased up to 200% without loss of content or functionality.</p>
+						</section>
+      
 		</section>
 
 		<section class="principle" id="operable">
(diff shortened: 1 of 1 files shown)


alastc:
So for those who struggle with reading the code, the SC text is currently:

> **Spacing:** Line spacing between paragraphs, blocks of text, graphics and embedded elements and objects and regions is at least 0.8em vertically and 0.5em horizontally and line spacing in cursive script languages have a line high of at least 150% unless a mechanism is available such that they can be increased up to 200% without loss of content or functionality. Level AA.

The and/or aspects are confusing me a bit, does this have the same meaning?
> Line spacing between paragraphs, blocks of text, graphics, embedded elements, objects and regions is at least 0.8em vertically and 0.5em horizontally, unless a mechanism is available such that they can be increased up to 200% without loss of content or functionality.
> 
> Note: and line spacing in cursive script languages have a line high of at least 150%

Also, "line spacing" is what you have between lines _within_ a paragraph (at least in CSS), but the wording implies you mean between paragraphs, is that right? In which case I'd just say "spacing".

For the units, it seems odd to say effectively: This is a good minimum spacing, but if you use less than that, allow it to be increased by 200%. 

If the spacing is 0.2em then 200% of that is 0.4em, which presumably is not enough?

The "mechanism is available" currently means "provide a mechanism", and for layout that will continue to be the case, even if personalisation is available.

Why not just say "...at least 0.8em vertically and 0.5em horizontally, unless a mechanism is available such that they can be increased to that size." (if a mechanism provided, I think the "content and functionality" bit is implied otherwise the mechanism hasn't worked.)

Also, for the EM unit it's probably best to link to the [CSS units definition](https://www.w3.org/TR/css3-values/#em). 

Have you considered how much variability there will be? For example, 0.5em of a 30px heading is 15px, but around a paragraph of 12px text the spacing of 0.8em would be 6px. So you would get 'slanted' pages where the headings are more indented than the text.

Perhaps it would be better to base it on [REM](https://www.w3.org/TR/css3-values/#rem), which is bound to the HTML root font-size? 

But then you could assume that people leave the default text size in their browser as 16px, so you might as well define it as 13 CSS pixels vertically and 8 CSS pixels horizontally...

awkawk:
Text Spacing SC adopted, but not using this PR so closing.

PR #114, opened 2017-02-09
Linearization Issue #58

DavidMacDonald:
There is a long discussion about this on my previous Pull request between Ian Pouncy, Patrick, and Alastair. See the discussion at issue #89  
See issue #58 for live links to these resources and a lively discussion that is both there and on a messed up pull request #89 which this replaces.

# SC Short title
Linearisation

# SC Text

Linearisation: A mechanism is available to view content as a single column, except for parts of the content where the spatial layout is essential to the function and understanding of the content.

# Description

Content can be viewed in a single column. All text is presented in single column format and the Line Length SC ensures that all lines of text word wrap to fit the available space. Data tables will have a standard tabular display, a two-dimensional matrix. Some user agents do not support active data such as form controls in reflow. In these cases the content author is not responsible for creating content that reflows.

Reflowed HTML
This good reflow of HTML was accomplished by a custom style sheet.

good PDF reflow

Properly reflowed PDF text.

bad PDF relow

Improper characters used by the author for spaces. This causes words to jam together with reflow. This authoring problem occurs frequently with PDF.

# Benefits

One important need of most people with low vision is to reconfigure a presentation from a normal two-dimensional page to a one-dimensional organization of data. This is not always the case, but it is a frequent need. For quick scanning the original structure may be useful to scan for recognizable items. This is usually done when the user knows a page and is looking for just one thing. In cases where careful examination of the page is necessary one column presentation is needed. The reasons are given below.

Continuous Reading: Many people, with and without disabilities, find it more difficult to read when they must scroll from the bottom of a column of text to the top of another column. With low vision the need is greater. To go from one column to next frequently requires paging up several screens to get from the bottom of one column to the top of the next. Additionally, using a small scrollbar and cursor is difficult with limited sight. This makes getting from the bottom of a column and finding the top of the next column more difficult. Reading flow degrades and comprehension suffers.
Reliable Searching: (visual navigation of the viewport) Searching for a specific item within a page of information is difficult for almost everyone with low vision. This is because the field of vision is limited either by enlargement requirements, or physical visual field loss due to tunnel vision. The ability to only scroll in one dimension vertical only or horizontal only but not two dimensions, dramatically simplifies this problem. A person can start at the top of the page and move element by element down the page until the bottom, and be certain that every item has been scanned. (TSBVI,Specific Eye Conditions, Corresponding Impact on Vision, And Related Educational Considerations)
Resizing Text on a Large Scale: Two column or more do not support large print text. Even if text stays within column boundaries, multi-column format will not provide enough space for text in large print. Even medium length words will not fit for sizes exceeding 250%. This causes excessive hyphenation and becomes distracting. This criterion is a necessary condition for useful text resize criterion for people with visual acuity worse than 20/70. Many people in this range require print size greater than 1/2 in. (1 cm.). This cannot be sustained with standard formats. content.
User Need: Users can set blocks of text in one continuous block, instead of in multiple columns.
Source: Accessibility Requirements for People with Low Vision, Section 3.2.2
Testability

HTML: Use the Easy Checks - A First Review of Web Accessibility document, Basic Structure check. Check that the sequence of text is in fact a correct reading sequence. Note: These tests remove all style. This means that items that the author did not intend to be seen become visible. It can be useful to display the pages side by side to compare what is intended to be visible and what is not. If this test passes the content can be restructured into an accessible format for low vision using CSS and JavaScript.
PDF: With a user agent that enables reflow, use the reflow option to see content reflowed into one column. No lines should be truncated. If this operation is successful the document passes.
Techniques

# Existing Relevant Techniques

G57: Ordering the content in a meaningful sequence (Note: Correct Reading Sequence and meaningful sequence are the same in this context.)
F1: Failure of Success Criterion 1.3.2 due to changing the meaning of content by positioning information with CSS
New Techniques

HTML: Write content so that the source order of perceivable elements defines a correct reading sequence. It is perfectly appropriate to include items that are not displayed to users in any convenient order. Such elements do not disrupt the correct reading sequence of the entire page. Make sure that elements that are not to be perceived are marked in a standard way like "display: none" or "visibility: hidden" so they can be ignored by a reformatting program. Test your document with no style in your user agent(s). See that it makes sense as a page and is fully operable. After this content is arranged and tested, positioning and other block formatting of the page. Use style sheets for positioning so that it can be removed and the resulting view is readable and operable.
HTML: Use media queries based on em units for reflow to single column. Reason: When the line length shrinks below a certain number of characters, reflow to a single column is triggered whether the reduced character count is caused by shortened lines or increased font size. This makes the reflow to single column accessible on more devices.
Fxx: Text becomes unreadable because of reflow.
Fxx: Failure of Success Criteria Text Colors and Reflow to Single Column due to inserting important information by using the background-image property of CSS.
Note: This is necessary because non-decorative images cannot be reflowed.
Related Information

# Articles

HTML Source Order vs CSS Display Order - Adrian Roselli
Email

Reflow is ready, and so is Line length - Wayne Dick, 15 November 2016
Terminology Question (LVTF thread) Wayne Dick, 1 November 2016
Latest Reflow - Wayne Dick, 31 October 2016
Reflow test case (Thread) - Alastair Campbell, 23 October 2016
Reflow Final Draft Wayne Dick, 4 October 2016
Reflow LVTF Thread 30 September 2016
I have made changes to Line Length and Reflow to Single Column - Wayne Dick, 20 August 2016
Requirements doc - Alastair Campbell, 18 February 2016
GitHub

Reflow to Single Column - WCAG 1.4.8 item 5 or New item LVTF GitHub SC Issue #7

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -368,6 +368,15 @@
 					<p>Instructions provided for understanding and operating content do not rely solely on sensory characteristics of components such as shape, size, visual location, orientation, or sound.</p>
 				</section>
 			</section>
+      
+      <section class="sc">
+					<h4>Linearisation</h4>
+					<p class="conformance-level">A</p>
+					<p class="change">New</p>
+					<p>A mechanism is available to view content as a single column, except for parts of the content where the spatial layout is essential to the function and understanding of the content.</p>
+				</section>
+			</section>
+     
 
 			<section class="guideline">
 				<h3>Distinguishable</h3>


awkawk:
Closing pull request as integrated into the FPWD.

Issue #14, opened 2016-09-18
Timeouts


lseeman:
-  [ SC for viewing](https://rawgit.com/w3c/wcag21/timeouts_ISSUE-14/guidelines/sc/21/timeouts.html) | [SC for editing](https://github.com/w3c/wcag21/blob/timeouts_ISSUE-14/guidelines/sc/21/timeouts.html)
-   [ understandable language definition for viewing](https://rawgit.com/w3c/wcag21/timeouts_ISSUE-14/guidelines/terms/21/understandable-language.html) | [understandable language definition for editing](https://github.com/w3c/wcag21/blob/timeouts_ISSUE-14/guidelines/terms/21/understandable-language.html)
-  [ SC in full draft guideline](https://rawgit.com/w3c/wcag21/timeouts_ISSUE-14/guidelines/#timeouts)

## Open issues and Surveys
Open issues: https://www.w3.org/WAI/GL/wiki/WCAG_2.1_SC_status#Issue_14_-_Timeouts
Surveys (Links to surveys require W3C Member access):
- [Results of Questionnaire Timeouts May 2, 2017](https://www.w3.org/2002/09/wbs/35422/Timeouts_Issue14/results) 

## SC Shortname
Timeouts

## SC Text

**For the current SC text, follow the links above. Below is the initially proposed text, which is now out of date**
Where content can time out, the content must also conform to all of the below:

- Loss of data: The user can easily return to the same point in a task, without data loss, for a period of at least one week as the default, or via a user-settable option available throughout the task. If the data will only be preserved for a limited time, the user is informed of the length of time that data are preserved at the start of the task.
- Timing adjustable: The function to turn off, adjust, or extend timing is controlled by a simple action, and is labeled with simple, understandable language.
- Aware: The user is informed of timeout limits at the start of the task, including the length of the warning.

## Suggestion for Priority Level

(A)

## Related Glossary additions or changes

- sensitive information
information that can put users at risk, or information a user wishes to protect
- essential
if removed, would fundamentally change the information or the functionality of content; and information and functionality cannot be achieved in another way that would conform
- safe standardized techniques
standardized by WCAG, other W3C groups, or by the platform, browser, or operating system, where user vulnerabilities are not exposed without informed consent (This will be a discussion, and is not yet finalized.)
- understandable language
conforms to all understandable-language success criteria

## What Principle and Guideline the SC falls within.

Principle 2, Guideline 2.2

This is an update to SC 2.2.1

## Description

[...]

PR #116, opened 2017-02-09
Timeouts - issue 14

lseeman:
from https://github.com/w3c/wcag21/issues/14

We should also try to make a AAA with the stuff that was taken out


michael-n-cooper:
Addressed at 40510f5487eb5a70c9ed33a2a026145adb45d368.

PR #117, opened 2017-02-09
Adding ASCII Art SC

awkawk:
This is not real!


awkawk:
just a demo PR so closing

PR #120, opened 2017-02-10
Resize content (Issue 77)

alastc:
Adding Resize content from #77 

## Proposed SC text

Content can be resized to 400% without loss of content or functionality, and without requiring two-dimensional scrolling except for parts of the content where fixed spatial layout is necessary to use or meaning.

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -493,7 +493,7 @@
                             <h4>Resize content</h4>
                             <p class="conformance-level">A</p>
                             <p class="change">New</p>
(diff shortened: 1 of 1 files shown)


mbgower:
There's something a wee bit awkward in there now. How about:

> Content can be resized to 400% without loss of content **or** functionality, and without requiring two-dimensional scrolling, except...

mbgower:
Actually, I'd also remove that comma after scrolling since the exception is about 2-d scrolling, right?

Content can be resized to 400% without loss of content or functionality, and without requiring two-dimensional scrolling except for parts of the content where fixed spatial layout is necessary to use or meaning.

awkawk:
merged to FPWD_review

PR #135, opened 2017-02-14
Plain Language (Minimum)

inclusiveThinking:
Plain Language (Minimum) #30


michael-n-cooper:
Addressed at 9a2f1570c9e31e431e705d65a425fb267d5dc116.

jtorresdelrey:
I am afraid that the definitions of Success Criterion 3.1.7 maybe too English-centric. For instance, labels and buttons may require infinitive or imperative forms in other languages such as Spanish, and those forms can also be easily testable. Use of modals like MUST and SHOULD would also be useful where. I would suggest concentrating on specific areas which are lexical, grammatical and discourse-analysis oriented, and then leave the specific techniques to each language (and to translators and localisers). For instance, use of X most common words in a language, or if a term is frequently used in a certain context but uncommon in the language  (e.g. pull request), there should be a clear definition using common words (but then we are in the 3.1.3 success criterion). As for grammar, use tenses which clearly and unambiguously allow (and require) to specify subject, object of the phrase, etc. As for discourse analysis or textual coherence, clearly refer to objects or actions which have previously or just before been presented or can be perceived in the vicinity or whose results can actually be tested before commiting the specif action, and so on.

Issue #70, opened 2016-11-29
Orientation


kwahlbin:
# Current versions of SC and Definitions
- [SC for viewing](https://rawgit.com/w3c/wcag21/orientation_ISSUE-70/guidelines/sc/21/orientation.html) | [SC for editing](https://github.com/w3c/wcag21/blob/orientation_ISSUE-70/guidelines/sc/21/orientation.html)
- [Essential definition for viewing](https://rawgit.com/w3c/wcag21/orientation_ISSUE-70/guidelines/terms/21/essential.html) | [Essential definition for editing](https://github.com/w3c/wcag21/blob/orientation_ISSUE-70/guidelines/terms/21/essential.html)
- [Display orientation definition for viewing](https://rawgit.com/w3c/wcag21/orientation_ISSUE-70/guidelines/terms/21/display-orientation.html) | [Display orientation definition for editing](https://github.com/w3c/wcag21/blob/orientation_ISSUE-70/guidelines/terms/21/display-orientation.html)
- [SC in full draft guideline](https://rawgit.com/w3c/wcag21/orientation_ISSUE-70/guidelines/#orientation)

# SC Shortname

Orientation

# SC Text

Content is not locked to a specific orientation, and functionality of the content is operable in all orientations, except where orientation is essential for use of the content.

# Suggestion for Priority Level (A/AA/AAA)

Level AA

# Related Glossary additions or changes

Orientation: portrait or landscape mode of the display
See the CSS Device Adaptation Module Level 1, the "orientation" descriptor [css-device-adapt-1].

# What Principle and Guideline the SC falls within.

Principle 2: Operable

New Guideline: Make content usable in device orientations.

# Description

Some mobile applications automatically set the screen to a particular display orientation (landscape or portrait) and expect that users will respond by rotating the mobile device to match. However, some users have their mobile devices mounted in a fixed orientation (e.g. on the arm of a power wheelchair). Therefore, mobile applications need to support both orientations by making sure content and functionality is available in each orientation. While the order of content and method of functionality may have differences the content and functionality is available. When a particular orientation is essential, the user needs to be advised of the orientation requirements.

## Examples of problems

Banking website locked in portrait mode
iOS home screen on the iPhone vs. iPad

# Benefits

Users with dexterity impairments, who have a mounted mobile device will be able to use the content in their fixed orientation

# Testability

- Turn the device into portrait orientation.
- Turn the device into landscape orientation
- Check all content and functionality is available in both orientations.

[...]

PR #140, opened 2017-02-16
ISSUE-70-ORIENTATION

marcjohlic:
#70  Created new Guideline “Orientation” (would not fit under any existing guidelines) under Operable.  Added new SC “Orientation”. Added additions and examples to glossary term “essential”.

This is for Issues #70 Orientation.  Ongoing discussion about whether a Note should be appended to this issue stating that authors must notify users when a change in orientation would cause a change in content (for example a nav bar in portrait mode may be changed to a hamburger menu in landscape mode).

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -509,6 +509,18 @@
 					<p><a>Section</a> headings are used to organize the content.</p>
 				</section>
 			</section>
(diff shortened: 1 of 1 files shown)


awkawk:
merged into FPWD_review

Issue #34, opened 2016-11-27
Reminders


lseeman:
<h2>SC Shortname</h2>
		        <p><strong> Reminders</strong></p>
		        <h2>SC Text</h2>
		        <p><br>
                <strong>Reminders:</strong> The user is able to  set a reminder for date and time sensitive events.  Reminders must be set only at the users request and the user must be able  to personalize the reminder method. Where a standard mechanism exists for the platform  or technologies, it must be used.</p>
		        <h2>Suggestion for Priority Level (A/AA/AAA)</h2>
		        <p>AA<br>
		        </p>
<h2>Related Glossary additions or changes</h2>
		        <p>Date and time sensitive event:<br>
	            Any event that has to be  completed by a certain time. The time constraints on such  an event may be defined by a calendar date and time or by the total elapsed  time.</p>
<h2>What Principle and Guideline the SC falls within.</h2>
		        <p>Principle 3, Guideline 3.3 - Input Assistance</p>
<h2>Description</h2>
		        <p>Managing events that are time bound (such as meetings, submitting a request by a  certain date or completing a form within a specified time period) can present  significant barriers to users with cognitive accessibility needs. The intent of  this Success Criterion is to ensure that when required, a user can set reminders and the delivery method of the reminder can be personalized by the user.<br>
For example:</p>
                <ul>
                  <li>A user may have difficulty processing and retaining time based information.</li>
                  <li>A user may have difficulty in sequencing time bound events.</li>
                  <li>A user's skills decreases when tired to such an extent that they have to stop a task. They may wish to reschedule the task.</li>
                </ul>
		        <h2>Benefits</h2>
		        <p>The benefit to users with cognitive accessibility needs is that they can manage appointments, deadlines and scedules. The ability  to set reminders can reduce the cognitive load associated when processing time bound  tasks. Time dependent activities  may be monitored and tracked by the user to  ensure that they are completed in a timely manner. This Success Criterion  addresses two broad classes of issues associated with this type of information:</p>
                <ul>
                  <li>If the user perceives the  activity to be too complex the user may decide to abandon the activity and  therefore be excluded from the information and/or services derived from the  completion of the activity.</li>
                  <li>If the activity relies on a  number of distinct events being carried out sequentially over an extended  period of time or if a single event must be completed by a specified date and  time then the likelihood of errors being made during the activity increases,  particularly for users with cognitive accessibility needs. Activities are often missed because the date and time is confused. </li>
                </ul>
[...]

PR #145, opened 2017-02-17
reminders

CityMouse:
Reminders #34 from COGA


awkawk:
Old Pull request, overcome by events as WG didn't adopt this SC.

PR #148, opened 2017-02-19
removing user testing

awkawk:


alastc:
I'm not sure lines 886 & 890 makes sense without the user testing wording? I.e. "shown to be easier" to whom?

At a stretch you could use "intended audience", but that it a fudge with its own questions.

PR #166, opened 2017-03-22
Fixes to 1 SC and 7 definitions that are in 2.1 but were coded as 2.0

awkawk:


awkawk:
arg. used the wrong branch.

PR #294, opened 2017-05-16
Update minimize-user-errors.html

mbgower:
- altered the title to Minimize User Input Errors
- added preface sentence
- added in the idea that a user can not use a separator character, and worked an absence of a separator character into the exception as well
- altered the second exception to make it a bit clearer
- added definitions
- ordered requirements and made exceptions unordered list


mbgower:
Note, I did not put the definitions into a definition list

PR #303, opened 2017-07-06
Adding "adapted" definition for issue 78

lauracarlson:
Per Greg Lowney's survey comment:
https://www.w3.org/2002/09/wbs/35422/AdaptingTextJuly6/results


steverep:
This change way over-complicates things in my opinion.  The proposal already says "by the user agent" and we're talking about style properties, so this doesn't add any substance.  This needs to be discussed on a call first.

lauracarlson:
Hi Steve,

Yes, let's discuss it on the call. Either including Greg's definition or not including it is fine with me.

PR #317, opened 2017-08-14
Create complex-gesture

detlevhfischer:
Hi,
I have provided definition text for the term "complex gesture" which is used in issue 61 Pointer gestures. This definition can probably still be honed and simplified. Sorry I have acted so late.

### DIFF
--- a/guidelines/terms/21/complex-gesture
+++ b/guidelines/terms/21/complex-gesture
@@ -1,0 +1,9 @@
+<dt class="proposed"><dfn>complex gesture</dfn></dt>
+<dd class="proposed">
+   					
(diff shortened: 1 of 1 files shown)


patrickhlauke:
i would change the term to "complex touch gesture", since it is exclusively defined as relating to touch

detlevhfischer:
Currently we have 'complex gesture' defined as 
A timed or multi-point gesture. Examples of a multi-pointer gesture include two-finger pinch/zoom
So the pinch gesture ist just one touch example - one could add a mouse drag-and-drop example to clarify it is not just touch

Sent from phone

Issue #67, opened 2016-11-29
Device sensors


kwahlbin:
## Current versions of SC and Definitions
- [SC for viewing](https://rawgit.com/w3c/wcag21/device-sensors_ISSUE-67/guidelines/sc/21/device-sensors.html) | [SC for editing](https://github.com/w3c/wcag21/blob/device-sensors_ISSUE-67/guidelines/sc/21/device-sensors.html)
- [Device Sensor definition for viewing](https://rawgit.com/w3c/wcag21/device-sensors_ISSUE-67/guidelines/terms/21/device-sensor.html) | [Device Sensor definition for editing](https://github.com/w3c/wcag21/blob/device-sensors_ISSUE-67/guidelines/terms/21/device-sensor.html)
- [SC in full draft guideline](https://rawgit.com/w3c/wcag21/device-sensors_ISSUE-67/guidelines/#device-sensors)

# SC Shortname

Device sensors

# SC Text
All functionality of the content can be operated without requiring specific device sensor information unless the device sensor is essential for the function and not using it would invalidate the activity.

# Suggestion for Priority Level (A/AA/AAA)

Level A

# Related Glossary additions or changes

**Editors´ Note:** The Working Group has identified questions of scope for this SC, and is exploring rewording to resolve this. The main question is about the relationship to indirect motion associated with operating a keyboard, pointer, or assistive technology. Draft new wording is available at https://github.com/w3c/wcag21/issues/67#issuecomment-325716721

**Device sensor:** A device sensor is a device component that detects and responds to some type of input from the physical environment.  Examples on mobile devices are the measurement of motion, orientation, and various environmental conditions.

# What Principle and Guideline the SC falls within.

Principle 2, *new* Guideline "Additional sensor inputs"

# Description

Devices may have sensors that act as inputs - e.g. tilt/orientation sensors on a mobile/tablet, allowing the user to control something by simply changing the orientation of the device in space. Not all devices may have these sensors. Even when the device has these sensors, the user may not be unable to operate these sensors (either not at all, or not precisely enough). Functionality must therefore be implemented in a way that other means are available to activate the function that do not rely on additional sensor input.

## Examples
* After text input in a field, shaking a device shows a dialog offering users to undo the input. A cancel button next to the text field offers the same functionality - activating it reverts to the content (if any) that was shown before the user input replaced it.

# Benefits

[...]

PR #378, opened 2017-09-11
Address scope and keyboard problems with Device Sensors SC

steverep:
SC proposal #67, Device Sensors, had a number of objections from the working group that have not been resolved.  This change is targeted to remove those objections and clear the editor's note in the draft which was agreed to as a compromise to continue work.

[Discussion of new language based on motion actuation](https://github.com/w3c/wcag21/issues/67#issuecomment-325485422) is available and changes are summarized here in this pull request:

1. Per many comments, the value is currently zero because the functionality must already be keyboard operable per 2.1.1, but what's really desired is not to have to plug in a keyboard.  So, the proposal instead requires UI components to perform the function.  Then there's a clear distinction since the UI component should be pointer operable on screen.
2. [Objections based on scope](https://lists.w3.org/Archives/Public/w3c-wai-gl/2017JulSep/0846.html), at least for me, could be cleared by talking about motion actuated functions rather than generically including all possible sensors.  This proposal does just that.
3. Not an objection, but I personally see much value in also making sure that such functions cannot be accidentally actuated by users with poor motor control.  This proposal requires the motion input to be able to be turned off.
4. A note is added simply to clarify direct motion sensation (e.g. accelerometers and gyroscopes) versus indirect motion like moving a mouse or pressing a keyboard key.


DavidMacDonald:
I agree it is much better and closer to what we want. 

However, the SC has to say what it means, can't rely on the Note to distinguish user motion in the air vs user motion in a gesture our mouse. The User is ALWAYS moving to create input. In speech they move their mouths, keyboard moves fingers, mouse/touch moves fingers wrists and arms. 

I would amend it to  something like "device motion or user motion where user does not contact the device"

steverep:
@DavidMacDonald, I agree, and this is the tricky part of the language that I struggle with.  I think your language to clarify the user motion might be an improvement, but "device motion" still has the same hiccup: keys and mouse buttons must move.  Without contacting the device also doesn't cover something like eye gaze operation.

For Silver, I'd like to see us introduce an "accessibility supported" concept for hardware-based input AT, and not just software AT working with web technologies.  In other words, require that any input method be relied upon only when there is a widespread way for people of all abilities to operate it, otherwise there should be an alternative.  That's obviously a much bigger topic though and I'd like to see something done to address it in 2.1.

steverep:
@DavidMacDonald, on second thought, the existing accessibility supported and assistive technology definitions work just fine if the APIs used to take motion-related input are considered web technologies just like anything else.  So how about just adding an exception whenever the input interface is accessibility supported like so:

**Motion Actuation:** Functionality which can be operated by device or user motion can also be operated by user interface components and can be disabled to prevent accidental actuation, unless:
- **Essential:** The motion is essential for the function and doing so would invalidate the activity; or
- **Accessibility Supported:** The input interface for the motion is accessibility supported.

The note could then be left there or just explained in Understanding.  What do you think?
Note: This criterion concerns input through sensors which respond directly to motions such as tilting, shaking, or panning. It is not intended to cover indirect motion associated with operating a keyboard, pointer, or assistive technology.

Issue #333, opened 2017-08-24
The Paciello Group comment on SC 1.4.14 Content on Hover or Focus


patrickhlauke:
"Visible trigger" references "essential", which (as noted in feedback for 2.6.1) is specific currently to exceptions for orientation. The definition of "essential" needs to either mention cases of "essential" in this scenario, or there should be separate "essential" definitions (for "essential orientation", "essential content", etc).

> "Either the additional content does not obscure any essential content within the triggering user interface component"

This seems unnecessarily complex to parse. Could this not be rewritten simpler to "...the additional content does not obscure essential content of the trigger"?

allanj-uaag:
lvtf

steverep:
> "Visible trigger" references "essential", which (as noted in feedback for 2.6.1) is specific currently to exceptions for orientation. The definition of "essential" needs to either mention cases of "essential" in this scenario, or there should be separate "essential" definitions (for "essential orientation", "essential content", etc).

I agree and I think this change to the WCAG 2.0 definition of essential was an oversight.  Many new and existing criteria, including this one, expect the 2.0 definition unchanged.  @awk and @joshueoconnor, can we prioritize fixing this to avoid more confusion?

> This seems unnecessarily complex to parse. Could this not be rewritten simpler to "...the additional content does not obscure essential content of the trigger"?

Again, I agree that this is wordy and adds the confusing implication that a trigger is always a UI component.  This was a change made in the last hours before the CFC on #75, and I objected to it.

DavidMacDonald:
+1

steverep:
@patrickhlauke, pending working group approval of pull request #569, the [proposed new language](http://rawgit.com/w3c/wcag21/hoverable-and-persistent/guidelines/#content-on-hover-or-focus) should address your issues.  Note the "essential" issue was already addressed, but this proposal would remove the word completely.

PR #393, opened 2017-09-22
Restore definition of essential to WCAG 2.0Restore definition of essential back to 2.0 and move examples to a no…

steverep:
This is a simple change to restore the definition of "essential" back to how it exists in WCAG 2.0.  Examples were added to this definition during the acceptance of the Orientation SC (#70), but it would be much more appropriate to add such examples in a note for that particular criterion since many other criteria use this definition.  The examples are shortened to a simple sentence in this change and can be expanded upon in Understanding.

This partly addresses #333 and will also help #372.


goodwitch:
+1  I support this change!

steverep:
@awkawk, the spelling error has been fixed and this is ready to merge.

PR #395, opened 2017-09-26
change foreground to #008400

ruoxiran:
The green text on green background (for the Editor's Notes) is not meeting color contrast. It is only 2.37 to 1. foreground color: #22bb22, background color: #e9fbe9
As Michael's suggestion: I add a style in guidelines.css, change it to foreground #008400/ background #e9fbe9


awkawk:
This is fixed as described in the PR specification, so closing.

Issue #375, opened 2017-09-08
"easily available" not in definition format


michael-n-cooper:
The definition for the term [easily available](https://w3c.github.io/wcag21/guidelines/#dfn-easily-available) does not fit into definition format we use from WCAG 2.0, where the first line of term should be a phrase that in principle could drop in place of the term within the SC text. The definition would come out as "one or more of the following are true:" viewed that way, which defers all substance to the subsequent list. I can't fix this one as an editorial change, so filing an issue.

joshueoconnor:
How many of our SCs referernce or use the term  [easily available](https://w3c.github.io/wcag21/guidelines/#dfn-easily-available) they may be mostly COGA SCs IIRC. @lseeman is this something you have a definition for?

detlevhfischer:
I find content and wording of the three bullets quite hard to understand.
if I parse, say, the second bullet as "easily available has the option to save or change the setting" that doesnt mit make sense.
I alsdo do not understand the meaning of "easily available can be set one time with as wide a scope as possible" - that doesn't even look grammatical.
The more fundamental issue is that it is rather soft term especially as it is now tied to three different conditions one of which has to apply. If it cannot be defined more succinctly, thi smay indicate a weakness of the term itself in the context of an SC.

detlevhfischer:
My proposal:
Remove the term "easily" in Success Criterion 2.2.7 Interruptions (Minimum). Instead of "A mechanism is easily available..." the SC text would then simply read:

> A mechanism is easily available to postpone and suppress interruptions and changes in content, unless they are initiated by the user or involve an emergency. 

We can then also remove the definition.
Rationale: How easy it is to avail yourself of a mechanism is not testable. The current definition does is unclear and does not help to decide / test unambiguously whether somthing is 'easily' available or not.

awkawk:
No longer in spec so closing

Issue #325, opened 2017-08-18
And/or in SC2.2.7 Accessible Authentication


guyhickling:
I see a small problem in ​the wording of SC2.2.7 (as shown in the 16 Aug 2017 draft):

> Essential steps of an authentication process, which rely upon recalling or transcribing information, have one of the following: 
 > - alternative essential steps, which do not rely upon recalling and transcribing information; or
> - an authentication-credentials reset process, which does not rely upon recalling and transcribing information

In the first line it says "recalling or transcribing", but in the two bullet points it replaces that "or" by "and".​ It should be "or" in all three cases.

The effect of saying "recalling and transcribing" in the bullet points is that anything that does not require BOTH recalling AND transcribing will be allowed.

For example a website could have a test requiring transcribing digits as the first alternative. It could then provide a second test of exactly the same kind but with different digits, as the alternative test. It would require transcribing, but not recalling, so doesn't get caught by "recalling and transcribing" of the first bullet point.

DavidMacDonald:
Proposed response:

Thanks for your comment, yes, that is a typo. It should be OR and not AND in the Two Bullet Points

PR #396, opened 2017-09-26
Addressed Issue 325 and editorial changes

DavidMacDonald:
Put in OR instead of AND for bullets to match the leading sentence. Updated the exception language to be more in alignment with current SC style. Removed the word EASY as per issue #375


DavidMacDonald:
Adresses Issue #325

awkawk:
Thanks David.

PR #398, opened 2017-09-26
Correct all uses of essential

steverep:
This request attempts to link all correct uses of the glossary term "essential" where links are missing, and also take the least obtrusive path to amending criteria where the term is used improperly outside of an exception.  This issue is discussed in #372.  A summary of changes is as follows:

## Correct Usage of Essential
Five new success criteria were found to be using the term properly in an exception, but were missing a link to the glossary:
* Graphics Contrast (exception only)
* Pointer Gestures
* Concurrent Input Mechanisms
* Device Sensors
* Accidental Activation

These were corrected by simply adding the link to make the usage clear.

## Incorrect Usage of Essential
The following criteria were found to be using "essential" incorrectly because it is not describing something that is exempt from the respective criterion.  Draft fixes for each are discussed separately.

### Graphics Contrast
This criterion restricts applicability to "graphical objects that are essential for understanding the content or functionality".  The draft fix here simply replaces "essential" with "necessary" as an appropriate synonym.

### User Interface Component Contrast
This criterion restricts its applicability to "essential visual identifiers".  Given that the notes and Understanding are going into detail about what that means and it is not tied to the definition, this draft just strikes the word "essential" from this phrase everywhere.  My logic is that saying "essential" is mostly redundant with the word "identifier", i.e. if something visual is identifying the type of component or its state, then it is always essential.

### Adapting Text
The criterion refers to "no loss of essential content or functionality".  The word seems to have been introduced by @awkawk in a [comment way back in March](https://github.com/w3c/wcag21/issues/78#issuecomment-289792275), but I could not find any rationale for its inclusion in the language (i.e. an example of content loss that would be acceptable).  Given this, and the fact that both Resize Text and Zoom Content refer to "loss of content or functionality" without using essential, this draft simply removes the word from this SC.

### Content on Hover or Focus
This criterion says that the additional content should not "obscure any essential content within the trigger".  The use here is meant to say that decorative content is okay to obscure, so this draft rewords the sentence to link to the definition of "pure decoration" instead.

### Accessible Authentication
This criterion restricts the applicability to "essential" steps, but it appears to me that this is really meant as a synonym for "required" steps.  The 2nd use in the bullet I believe is unnecessary to the SC.  So, this draft simply replaces "essential" with "required" to correct the former misuse and delete the latter use.

### Animation from Interactions
This criterion is actually using the term correctly as an exception to restrict applicability to "non-essential animations", but it's an awkward read.  This draft just moves the clause to a clear exception at the end.

### No Timing  (not addressed via this request)
Yes, an existing AAA criterion is using "essential" incorrectly.  However, it is clear to me from the SC that this is simply meant to be a synonym for "required", and the Understanding content even just says required and gives no rationale for using "essential".  I recommend we replace "essential" with "required" as WCAG 2.0 errata and merge into 2.1.


steverep:
Most of these changes are now obsolete and the issue is much smaller in the current working draft.  Closing to re-evaluate.

PR #500, opened 2017-10-09
Graphics contrast

alastc:
Initial re-drafting of graphics contrast.


alastc:
NB: I still have some images to add, getting a little design help with that to avoid any copyright issues.

Issue #501, opened 2017-10-09
Comment on 2.4.11 Character Key Shortcuts


jnurthen:
Some clarification still needs to be made as to where a single key shortcut is acceptable and where it is not. The understanding document attempts to do this but IMO it is incomplete.

For example a SELECT does not need to be opened in order to use the single character shortcuts. It merely needs to be focused. There are many similar controls which appear on the web today - both standard and non-standard controls where the single character shortcut helps usability for keyboard users.

It must be clarified that these controls can use single key shortcuts when they have focus.

DavidMacDonald:
Proposed Responses

We've created a pull request https://github.com/w3c/wcag21/pull/512 with an exception for when a character key shortcut is only active on focus rather than on page load.

joshueoconnor:
Survey https://www.w3.org/2002/09/wbs/35422/Oct_17th_agwg/results

awkawk:
RESOLUTION: Add "unless the keyboard shortcut is only active when the user interface component has focus"

"If a keyboard shortcut consisting entirely of one or more character keys is implemented by the content, then a mechanism is available to turn it off or to remap it to a shortcut that uses at least one non-character key, unless the keyboard shortcut for a user interface component is only active when that component has focus."

PR #512, opened 2017-10-10
Added exception for issue #501

DavidMacDonald:
Added exception for when an app enables a character key when a control has focus, but it is not active at other times.


steverep:
The keyboard shortcut, not the character key, should be the subject of the exception.  Also, change control to user interface component.  I recommend changing to:

> ..., unless the keyboard shortcut is only active when a particular user interface component has focus.

PS - Again this should be going into master if it gets WG approval.

DavidMacDonald:
Updated exception as per @steverep friendly amendment, and Pulled against the Master branch.

Issue #513, opened 2017-10-10
Grammar fix on User Interface components SC


DavidMacDonald:
It says:
> ```Examples of <a>essential</a> visual identifiers of <a>user interface components</a> may include (a border, edge, or icon), current value (such as non-text visual indication ...```

There is a bracket after includes. I think the brackets are examples of the thing before it, but there is nothing before it.

I propose adding the term "visual identifiers"

> ```Examples of <a>essential</a> visual identifiers of <a>user interface components</a> may include visual identifiers (a border, edge, or icon), current value (such as non-text visual indication ...```

Added pull request  #514

May need another tweak because the line ends with "other essential visual indication". This could probably be deleted but should have some discussion first. 

Note: the PR references Issue 506, and that is what the branch is called... but it should be called Resolve issue 513. My bad.

joshueoconnor:
Surveyed https://www.w3.org/2002/09/wbs/35422/Oct_31_2017/

awkawk:
closing as pull request approved by WG

PR #514, opened 2017-10-10
resolve Issue #513

DavidMacDonald:
Added a term before the first brackets to make them make grammatical sense. May need another tweak because the line ends with "other essential visual indication". This could probably be deleted but should have some discussion first.

### DIFF
--- a/guidelines/sc/21/change-of-content.html
+++ b/guidelines/sc/21/change-of-content.html
@@ -14,8 +14,7 @@
    </p>
    					
    <ul>      						
-      <li>There is a programmatically determined relationship between the new content and the control
-         that triggers it;
+      <li>There is a programmatically determined relationship between the new content and the triggering mechanism;
       </li>						
       <li>The user has been advised of the change of content before or as a result of using the control;</li>
       <li>The change of content is not a result of a user action AND not related to the <a>primary purpose of the page</a>.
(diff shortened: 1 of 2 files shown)


goodwitch:
Hi David...I think there is some redundancy...and I'd suggest a few changes.  With this pull request it would read:  "Examples of _essential_ **visual identifiers** of user interface components may include **visual identifiers** (a border, edge, or icon), current value (such as non-text visual indication of aria-valuenow on a slider) and current state (such as selection indicator, focus indicator) or other essential visual indication (which do not rely on color alone)."

The double use of "visual identifiers" in the same sentence is questionable to me.  I think we should change the 2nd occurrence of "visual identifiers".  Maybe:

Examples of  **visual identifiers** of user interface components may include **informative non-text content** (a border, edge, or icon).

I also think the word "essential" is redundant.  If something is a "visual identifier"...and you take it away...then that thing is no longer identified.  We can avoid the problems with essential definition from WCAG 2.0 by either removing it (or replacing it with the word "necessary").

DavidMacDonald:
@goodwitch I've made the changes, removing "essential" and replacing the first instance of visual identifiers with "informative non-text content"

awkawk:
This was going to be an easy editorial edit, but with the additional changes we will need to survey it. That's ok, but just so you know...

DavidMacDonald:
note the any references to issue 506 should be issue 513 for this fix. One the branch is merged and deleted there will be no more references to 506. If there is too much confusion I can blow away the branch and rename it, hope not to have to do that.

awkawk:
closing after implementing in the master branch. Was easier to do manually than to redirect the pull request to the right branch.

PR #524, opened 2017-10-11
First pass at contextual information understanding doc.

chriscm2006:


lseeman:
Hi

We seem to have a miscomuncation . we have an understanding section at https://docs.google.com/document/d/13hmoaVU563kTio1EZD5mbNxcc0k924qVdZZwWckcbu0/edit


can you include it

All the best

Lisa

chriscm2006:
Hi Lisa, 

Yes thanks much for that work, but there is no misunderstanding. Jake and I read through your content extensively before working on the respective "Purpose of Controls" and "Contextual Information" intent documents. The problem is the one you have written for the two are essentially identical, which is not clear. Certainly the two are certainly highly linked, but they are not identical. I have pulled the parts out of that document that match up with "Contextual Information" and Jake is doing (probably already has done) the same for Purpose of Controls.

What you have was great info, it just needed to be separated into the parts to make the distinction between the two SCs clear.  If you think there's something missing feel free to comment, but make sure you do so within the context of viewing work that @jake-abma has done on Purpose of Controls.

Chris

awkawk:
This pull request refers to an old version of the SC that eventually turned into "Identify Purpose" - closing since it isn't relevant now.

PR #543, opened 2017-10-25
Added metadata paragraph and example

DavidMacDonald:
to the conformance claim section as per
https://github.com/daisy/epub-revision-a11y/wiki/WCAG-Discovery-Metadata-Proposal

### DIFF
--- a/understanding/20/conformance.html
+++ b/understanding/20/conformance.html
@@ -226,7 +226,7 @@
          descriptions</span> are also provided for images that require more 
        than simple alternate text. All content is available in text, which 
        can be accessed by assistive technology.</p>
-</div><code>
+</div></code>
     
   <p>Sometimes, one may want to make a claim for just the content that was added after
          a certain date. Or, one may want to claim WCAG 1.0 conformance for content up to a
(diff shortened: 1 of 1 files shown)


DavidMacDonald:
Note:  I've put the example inline, however later it might make sense to create a technique with this example, and then just the link will be in the document. This could be considered, but I think it reduce the prominence of the epub example,

understanding/20/conformance.html
```diff
@@ -193,10 +193,42 @@ <h3>Understanding Requirement 5</h3>
...
+    </ul>
+    <p>This page includes both <span property="accessMode" content="textual">text</span>
+       and <span property="accessMode" content="visual">images</span>.
+       <span property="accessibilityFeature" content="alternativeText">Alternative 
+       text</span> is included for all image content and <span 
+         property="accessibilityFeature" content="longDescription">long 
+         descriptions</span> are also provided for images that require more 
+       than simple alternate text. All content is available in text, which 
+       can be accessed by assistive technology.</p>
+</div><code>
```
alastc: I assume this should be a closing tag?

awkawk: yes, fixed

awkawk:
WG agreed to merge this on May 8 2018. Problematic to merge with Github since the conformance.html page location changed and it was easier to copy and paste the change into the file directly. Closing this PR but the content was merged.

PR #545, opened 2017-11-01
Understanding cav edits to Understanding 

awkawk:


awkawk:
Changes to Understanding to address #306

PR #566, opened 2017-11-06
Updated Zoom for grammar and clarity

DavidMacDonald:


alastc:
Hi David,

I did a fairly large overhaul of the zoom understanding doc:
https://github.com/w3c/wcag21/blob/zoom-content/understanding/21/zoom-content.html 

I think the bits you've proposed are no longer there, but I had similar ideas about adjusting some of the text... sorry, I should have accepted the merge before starting.

-Alastair

PR #584, opened 2017-11-17
Pointer gestures

detlevhfischer:
I have updated the text of the SC "2.5.1 Pointer Gestures" and the corresponding Understanding text to reflect the need to allow for long press and double click. I have also added a third example.
https://github.com/w3c/wcag21/blob/pointer-gestures/understanding/21/pointer-gestures.html


michael-n-cooper:
I am manually merging working branches and resolving conflicts

PR #591, opened 2017-11-27
Adapting text changes

awkawk:


lauracarlson:
Editorial: Maybe consider linking to the definition of "blocks of text" in the editor's note?

PR #594, opened 2017-11-30
Status changes

DavidMacDonald:
Created diff "change of Content" file
Created clean copy of status-changes file
created definition


awkawk:
should the change of content HTML file be deleted?

DavidMacDonald:
Yes

awkawk:
Can be viewed at http://rawgit.com/w3c/wcag21/status-changes/guidelines/index.html#status-changes

steverep:
I've scoped this to markup languages per @awkawk's request, and added @alexswli's requested editor's note.  I think this should be ready for CFC.

PR #607, opened 2017-12-07
Additional sensor inputs

detlevhfischer:
I have added draft text for guideline 2.6 Additional sensor inputs; will work on the understanding doc now.


michael-n-cooper:
This branch was intended for work on Understanding content alone. There were commits that affected the guidelines file, and a rename of an understanding file for device sensors to motion actuation that should have happened in the working branch for that SC. So instead I spun out a new branch, reverted non-Understanding commits, and merged that. https://github.com/w3c/wcag21/commit/df8afe035ab8da92bee97ebb44d2bb56e6e380ba I'm leaving this branch alone so we can later come back to the guidelines edits, but closing this pull request.

steverep:
Are we bypassing working group consensus here?

michael-n-cooper:
We discussed the process in [today's call](https://www.w3.org/2017/12/07-ag-minutes#item02) with reference to a process we agreed to by CfC some months ago. If you have a specific objection to this merge please raise it, but I am just trying to merge Understanding branches that I have been told in some fashion are ready to go, within this process we accepted.

steverep:
I did not realize you reverted the non-understanding changes, so I have no objection at this time.

SuzanneTaylor:
"Provide alternatives for user input via device sensors." Doesn't cover 2.6.2 Maybe something like "Make functionality available without requiring real world motion or positioning." This covers not requiring users to, for example, hold the phone up high, position the phone in landscape, move the phone in a specific way, etc
OR 
"Require the least amount of real world motion or positioning needed for functionality." This might leave the door open for some SC that require adjustments to the sensor-based functionality to make it easier to use, as well as requiring alternatives. For example, for inputs for AR and VR, wheel chair users (and many others) can often be accommodated by just moving the target objects downwards. If the requirement is always to just provide a full alternative that doesn't require motion, the full experience for wheelchair users (and many others) might be unnecessarily less engaging. And, once new AT appears for this, even more people with mobility impairment might also prefer that type of accommodation, rather than simple replacement.  (AT down-the-road might remap an app's understanding of where the phone is to the position of a virtual phone, but control of that virtual phone might easiest when restricted to some extent.) It's kind of similar to providing alt-text for small blurry images. For some users, a better image would be best, but the guidelines require alt-text. However, if AR and VR become significant, this situation might be the difference between participation and exclusion from entire categories of software, which is not the case for alt-text/images! Techniques in addition to replacement should at least be mentioned as optional best practices...

SuzanneTaylor:
I'm not sure the best place to leave this comment, but I had one last thought on WCAG 2.1 
It is just an idea for the understanding documentation: In practice with this, I've found that the UI controls that provide an alternative to sensor-based input also double as a feature for your application's power users. If the app requires a user to shake the phone - that might be fun the first few times, but a power user likely doesn't want to do this over and over.  Also, there is a study somewhere that shows that for apps where you use the device as a steering wheel, a large percent of people without mobility impairment chose that option - e.g. a person on a train probably just wants to click and not be moving their phone around a lot, bumping into the person next to them. If the link to study would be helpful, I can try to find it, just ping me on twitter @ThingsEGames

PR #610, opened 2017-12-07
Harmonized with official group list

awkawk:


michael-n-cooper:
I expect there was a need to catch up the list to new joins, and I see some alphabetical sorting errors had crept in as well. However, one attribute of the list was that it showed *active* participants, not people who joined the group and never did anything. That's tricky because I tried to capture people who are active in task forces even if not the main group, but there are people who as far as I can tell aren't active in any way so aren't included. While we can discuss I'd like to keep that attribute of the list, so I'm not sure if I want to accept this pull request.

detlevhfischer:
SSB Bart Group is now Level Access.

PR #633, opened 2017-12-15
Proposed wording for additional sensor inputs GL

michael-n-cooper:
taken from
https://github.com/w3c/wcag21/commit/fb1692ba6c1d91f2616f11cbb0c68f9326780c0e
by @detlevhfischer but in an appropriate review branch


steverep:
I thought we had also agreed to at least drop "Additional" from the handle?  I would prefer a very simple handle like "Sensors".

PR #643, opened 2017-12-20
update per issue 618

awkawk:


awkawk:
closing without merging since the SC was removed from editor's draft.

Issue #634, opened 2017-12-15
single pointer definition


detlevhfischer:
Currently, the new definition term 'single pointer'
https://www.w3.org/TR/WCAG21/#dfn-single-pointer
is defined as:
> one point of contact with the screen

My concern is that this does not make it clear whether this means an untimed action on one point on the display (x/y coord), or whether it would also include long press and double click / double taptap.

@mbgower had concerns in an earlier comment that limiting 'single pointer' to just one untimed click or tap would be too restrictive, basically disallowing long press and double click which authors might want to port over to web apps from general software.
While the term 'untimed' has been removed in the latest SC language for 2.6.1 Pointer Gestures
https://www.w3.org/TR/WCAG21/#x2-5-1-pointer-gestures
it is now unclear what "can be operated with a single pointer" exactly means. Does it include long press / double click? Can we leave that to the Understanding doc or do we need to amend the definition of 'Single Pointer' and/or the text of 2.6.1 'Pointer Gestures' to clarify that?

awkawk:
@steverep can we assign making a response to this for you?

steverep:
The term "single pointer" needs to include single, double, triple, etc. clicks, long presses, and single point gestures, otherwise the most recent version of the [Pointer Cancellation SC](https://www.w3.org/TR/WCAG21/#pointer-cancellation) becomes largely worthless.  My interpretation of the definition was basically only an exclusion for multiple points of contact.

I presume this is the desired definition for both SC, and I agree it could use some additional clarity.

detlevhfischer:
I have created a pull request for an updated single pointer definition: https://github.com/w3c/wcag21/pull/664
@steverep: With reference to the comment in issue #663 by @mapluke, I think 'single pointer' would not just exclude multi-point, but also path based-gestures (and the SC text of 2.5.1 Pointer Gestures should be extended to make this clear), in other words, single pointer activation happens at the same screen position.

awkawk:
(Official WG Response)
Thanks for the comment. We agree that the definition needs to be clarified. We are changing the definition to:

"Pointer input that operates with one point of contact with the screen, including single taps and clicks, double-taps and clicks, long presses, and path-based gestures"

PR #664, opened 2018-01-09
Update single-pointer.html

detlevhfischer:
As follow-up to my issue  [#634](https://github.com/w3c/wcag21/issues/634) and Steve's recommendation I have created a branch for single-pointer definition (that didn't seem to exist so far) and drafted an alternative defintion for "single pointer" that is referenced both in 2.5.1 Pointer Gestures and 2.5.2 Pointer Cancellation.

### DIFF
--- a/guidelines/terms/21/single-pointer.html
+++ b/guidelines/terms/21/single-pointer.html
@@ -3,6 +3,6 @@
    					
    <p class="change">new</p>
    					
-   <p>A single pointer describes a means of input that operates on one particular screen location. The input can be indirect (a mouse, joystick or trackpad cursor) or direct (a stylus or finger on a touch screen). The operation through single pointer includes single taps and clicks, double taps and clicks as well as and long presses. It excludes path-based and multi-point gestures.</p>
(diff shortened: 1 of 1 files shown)


awkawk:
Suggest a change in form to align with other definitions:
Pointer input that operates on one screen location, including single taps and clicks, double-taps and clicks, and long pressed.

The input can be indirect (a mouse, joystick or trackpad cursor) or direct (a stylus or finger on a touch screen). It excludes path-based and multi-point gestures.

detlevhfischer:
Thanks, sounds good to me ('long presses' instead of 'long pressed'?)

steverep:
@detlevhfischer, restricting the definition to one screen location is not going to work with the Pointer Cancellation SC.  That SC was specifically designed to include single point gestures, drag and drop, etc.  We might want to reword the Pointer Gestures SC instead.

detlevhfischer:
@steverep  
How about " 2.5.1 Pointer Gestures: All functionality which uses multipoint or path-based gestures for operation can be operated with a single pointer at one screen location..."?
The single pointer definition would then have to change to something like "Pointer input with a single point of contact, including single taps and clicks, double-taps and clicks, and long presses." And we would have to take out the bit "It excludes path-based (...) gestures" if this is what is needed for the undo actions in Pointer cancellation.
Hm, I don't like it. It feels like 'single pointer' is not going to work well for both SCs. But it is the best I can come up with right now... I hate to mess this up again shortly before the door closes.

awkawk:
Old Pull request, overcome by events..

Issue #637, opened 2017-12-17
Mentioning RFC 2119 keywords on the body text


momdo:
Most of specs not only refer RFC 2119 as normative reference but also state the RFC 2119 keywords like this:

> The key words MAY, MUST, MUST NOT, SHOULD, and SHOULD NOT are to be interpreted as described in [RFC2119].
https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#conformance

I think that WCAG 2.1 needs to do so.

awkawk:
Thank you for the comment. We will add the notice as done in ARIA. Please note that these are the correct interpretations for these terms even without this sentence since the document indicates that it is governed by the process document and the process document makes this same claim.

awkawk:
Changes implemented - see at http://rawgit.com/w3c/wcag21/master/guidelines/index.html#conformance

PR #726, opened 2018-01-13
Clarify interpretation of conformance

### LINKED ISSUE #374: "normative" and "informative" terms not referenced
Inherited from WCAG 2.0 - the glossary contains definitions of the terms "normative" and "informative", but those terms aren't referenced anywhere in the spec. It was probably meant to be referenced from the conformance section, but somehow that didn't happen. Do we want to add references to those terms, or just remove the terms? 

Note the Respec script uses the terms "normative" and "non-normative" rather than "normative" and "informative", to be aware of when we fix this.

michael-n-cooper:
Add a section in conformance to describe how the guidelines are impacted
by conforamance and reference the terms "normative" and "informative".
Closes #374.

### DIFF
--- a/guidelines/index.html
+++ b/guidelines/index.html
@@ -330,9 +330,16 @@
             </section>
 
         </section>
-        <section id="conformance">
-            <!--<h1>Conformance</h1>-->
+        <section>
+            <h1>Conformance</h1>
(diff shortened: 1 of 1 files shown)


michael-n-cooper:
This pull request would also close #637

PR #727, opened 2018-01-14
Added definition for Animation

alastc:
Based on comment #697. Not sure if it is ok to change the first word of SC text "Animations" to the singluar to match the definition?


awkawk:
@alastc Did you make a new branch off Master before making the changes or did you start from the "animation-from-interaction" branch that was already present?

awkawk:
Also:
1) Please leave the SC text as it was, that is what the data-lt attribute in the definition is for
2) The definition seems to be for character keys

alastc:
Ach, just seen that I was trying to merge into master, that wasn't intentional.

I haven't edited for a while so I merged all the things. (W3C-master > alastc-master, W3C-animations > alastc-animations, and just about every connotation I could find that went from W3C to my fork.

I think if we cancel this one, I'll undo the change to the SC text and do a PR to the animations branch.

alastc:
Oh dear, if I merge my animations branch to the W3C one, it brings over all the changes from master, around 190, which is worse than this one! Doh, not sure how to back though out.

alastc:
Ok, done those. It also brings in a minor update to the understanding I was doing on the same branch, sorry.

Is it ok that the SC text is "Animations" but the definition is "animation", singluar?

alastc:
@michael-n-cooper I've done those apart frmo the readme one, that isn't just backing out, it came from merging master into my branch (trying to avoid conflicts). I'm not sure how to exclude that, I can't undo it as such.

detlevhfischer:
My one concern would be whether the definition addresses the point of one-step-only changes which would bring like user-initiated changes of size, added outline etc done for highlighting focus into the realm of this SC. Is that intended. I think not. I think Patrick made this point, and it seems open in this definition.

alastc:
Hi Detlev,

From what we know of the vestibular triggers, those types of changes aren't a trigger, it requires a visible transition. If there is no transition, it isn't covered by this definition.

detlevhfischer:
Ok I see that addition of steps between states implies that a third step is added . so if someone would grow an 2px focus outline via a 1 pixel step) it would technically fall under animation and likely still not cause issues. This may be nitpicking and I am certainly not prressing this point, just saying...

The one thing that seems a bit odd in the revised definition in CfC of issue 697 is the bit 'For example', because the first thing that follows is not an example for the def text. I would just drop "For example". One could actually add the common focus outline case that people may worry about for clarification, e.g.:
For example, an element which appears instantly in one frame @@@ or as a result if focusing@@@  is not using animation. This is just an idea and certainly not critical, not meant to slow things down..

patrickhlauke:
I'd still voice a slight concern that this tacitly only counts movement and/or size change as animation, without any mention of color, opacity, or other types of animation. Would be good to clarify somewhere (normatively, ideally, or non-normatively) that the scope of the SC is limited to those types of animation.

awkawk:
I have been making the edits of this sort by making a new branch off of the current Master and then making the changes. This minimizes the file collisions substantially.

mraccess77:
@patrickhlauke.  If I recall correctly we were trying to target specific types of animation such as parallax scrolling      So I believe it was limited based on information as we knew that were problematic.    As for the definition of animation even if  there were no changes in a frame - animation between frames could be an issue but I don’t think weee wanted to tackle all of that for this round.

alastc:
@patrickhlauke:
> I'd still voice a slight concern that this tacitly only counts movement and/or size change as animation, without any mention of color, opacity, or other types of animation. Would be good to clarify somewhere (normatively, ideally, or non-normatively) that the scope of the SC is limited to those types of animation.

Working from comments at TPAC, the intent seemed to be that the SC is not scoped to purely movement. Although the main use-case is movement/scaling, the thought was that it would help otehrs if it was all animation. Also, a definition of animation should not be scoped to just movement either. Do you think the new text excludes other types of animation?

@michael-n-cooper I'll have to re-do this tonight, unless we can use Andrew's version. I made the mistake of merging W3C master into my animations branch, thinking it would be upto date and reduce conflicts. It had the opposite effect.

patrickhlauke:
> Also, a definition of animation should not be scoped to just movement either. Do you think the new text excludes other types of animation?

no I think there's no evidence (unless I missed it) that animations other than motion/size animations cause problems to users with vestibular issues. so this SC would require authors to, for instance, use the reduced *motion* media query as a hint to avoid *any* kind of animation (like a subtle fade in color or similar). it would fail things that aren't actually problems.

patrickhlauke:
xref https://webkit.org/blog/7551/responsive-design-for-motion/

alastc:
In which case we're flip-flopping on this, as we had scoped it to motion & scaling animations, and moved back general animation (prior to the comments period).

As a counter-point, is blurring a movement? Also, imaging scrolling down a page quickly and the colour flips back and forth at over 3Hz... 

If this was at AA I'd agree it should be more specific, but the group thought that making general (at AAA) was the best way to go.

patrickhlauke:
> As a counter-point, is blurring a movement? Also, imaging scrolling down a page quickly and the colour flips back and forth at over 3Hz...

then we're probably back at the problem of wanting to say "significant" animation or similar, which is impossible to normatively quantify/define.

mbgower:
> then we're probably back at the problem of wanting to say "significant" animation or similar, which is impossible to normatively quantify/define.

We quantified this previously, however that was removed because we had insufficient data on what combination of length and size was a predictable trigger. So I wouldn't say it is necessarily impossible to quantify, but rather we lack data to define.

alastc:
Proposal sent to the list:

SC: “Motion animation triggered by user interaction can be disabled, unless the animation is essential to the functionality or the information being conveyed.”

Def: “Motion animation: addition of steps between states to create the illusion of movement and/or to give a sense of movement.
For example, an element which moves into place or changes size while appearing is considered to be animated. An element which appears instantly in one frame is not using animation. Motion animation does not include changes of color, blurring or opacity.”

The above change covers the original intended scenario (vestibular disorders), if we also want to cover colour and other types of animation (e.g. for some cognitive folk) the definition should be changed to say something like:
“addition of steps between states to create the illusion of movement and/or to draw attention.” With some additional examples.

alastc:
Closing as I've created a cleaner pull request.

PR #729, opened 2018-01-15
exception added for issues 657/677/685

awkawk:


lauracarlson:
@awkawk thanks for this pull request. 

It doesn't seem as though "Spacing underneath paragraphs" was edited to "Spacing following paragraphs" as we said we would change for [issues 657 and 677 and 685](https://www.w3.org/WAI/GL/wiki/Draft_Responses_to_Dec_WD_Issues#657_and_677_and_685). Could that please be done?

PR #734, opened 2018-01-15
Issue 616, renaming Graphics-contrast to non-text-contrast

alastc:
Not sure how easy it is to rename a branch here, but this is a change to the name for graphics contrast (i.e. the SC text and understanding doc) from issue 616

There are quite a few changes to the understanding doc as i've been working on it, but the main thing is the SC name change.


joshueoconnor:
@alastc Rather than renaming the branch, may be best to create a new branch with your preferred name, check that out, make the changes you want and then submit a PR against it.

awkawk:
I'm not sure why you need to rename the branch? Any old branch name works. You do need to edit the index.html file to change the reference from graphics-contrast.html

alastc:
Ah, thinking about it, the branch name in the W3C repo is entirely up to you, I'm merging in from a fresh branch on my repo so it lines up better. Index updated.

PR #738, opened 2018-01-16
Animation from interactions motion

alastc:
This is replacing #727 with an updated definition. Less merging issues to.


awkawk:
Will we ever use the term "motion animations"? If not, you can remove the data-lt= bit

alastc:
Err, I don't know what data-lt actually does, so didn't want to mess with it!

Issue #711, opened 2018-01-12
Intopia Feedback – SC 1.4.13 Content on Hover or Focus


AndrewArch:
1.	We think that ‘hoverable’ needs to be expressed more plainly (rather than relying on the explanation in Understanding)

awkawk:
Thanks you for your comment. The deadline for comments has passed, but the Working Group will make efforts to consider your comment and any suggestions for improvements.

awkawk:
(Official WG Response)
Thank you for your comment. While the intent and requirement for the Hoverable condition has always been the same, we've struggled to state it as simply as possible in the SC. We have taken your suggestion to state it more plainly and edited it to read:

> **Hoverable:** If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

Issue #695, opened 2018-01-12
The Paciello Group comment on 1.4.13 Content on Hover or Focus


patrickhlauke:
> Hoverable: If pointer hover can trigger the additional content, then the pointer can be moved to hover the additional content;

Perhaps remove the second "hover", as it makes it sound unnecessarily confusing, leaving just "then the pointer can be moved to the additional content"?

awkawk:
Thanks you for your comment. The deadline for comments has passed, but the Working Group will make efforts to consider your comment and any suggestions for improvements.

steverep:
Hi @patrickhlauke.  Several folks suggested just changing it to "hover over" instead.  I am working on a pull request for that and other changes now.  I assume that would address your issue.

patrickhlauke:
not quite sure. would that preclude the user then being able to also click or similar actions other than hover? it's not a big deal either way, just seemed unnecessarily complex to talk about being able to hover over what appeared after hovering.

alternatively, expanding the above suggestion, to make absolutely clear what i assume is meant "then the pointer can be moved to the additional content without the content disappearing" or similar.

awkawk:
(Official WG response)
Thank you for your comment. While the intent and requirement for the Hoverable condition has always been the same, we've struggled to state it as simply as possible in the SC. We have taken your suggestion to state it more plainly and edited it to read:

> **Hoverable:** If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

Issue #650, opened 2017-12-28
Success Criterion 1.4.13 Content on Hover or Focus


milk54:
Is the tab menu included in this success criterion?
The accessible tab menu displays the content when it receives the focus.
Of course, Tab menu displays the content by clicking. So Do not include those that work on clicks on this success criterion, is not it?
(Only when hover or focus only？ )

I do not know how to techniques this Success Criterion, so I would like you to provide many concrete examples for easy understanding.

awkawk:
(Official WG Response)
Thank you for your comment. Tabbed interfaces and similar selections, such as a drop-down box, that may cause additional content to become visible on focus/selection were not intended to be covered by this criterion. Rather, the criterion was designed for content that also becomes hidden when focus (or hover) is removed. We agree the criterion text did not reflect this well and made a change so that it now reads:

> Where receiving and removing pointer hover or keyboard focus triggers additional content to become visible and hidden, respectively, the following are true: ...

This scopes out tabbed interfaces because the removal of focus does not then hide the content because it is still selected. In this case, it might seem like the content is removed when the tab loses focus, but the content is actually changed when a different item receives the focus. We hope this addresses your concerns.

Regarding techniques, the working group will be working hard on the supporting understanding and technique materials in the next few months.

milk54:
I understand. Thank you.

PR #739, opened 2018-01-17
Edits to Content on Hover or Focus

steverep:
Changes for this pull request can be viewed at: http://rawgit.com/w3c/wcag21/hover-focus-edits/guidelines/#content-on-hover-or-focus

The following changes are made:
1. Revision to Hoverable condition to use more plain language (issues #695 and #711)
2. Revised initial sentence to scope to additional content which also disappears on hover or focus.  This is needed to scope out content that may appear on focus such as in a tab list, drop down box, etc. (issue #650)
3. Adds an exception to the Dismissable condition when the additional content doesn't obscure or replace anything.  This addresses comments on mailing list regarding things like help content inserted inline when an input field has focus.

### DIFF
--- a/guidelines/sc/21/content-on-hover-or-focus.html
+++ b/guidelines/sc/21/content-on-hover-or-focus.html
@@ -11,7 +11,7 @@
   <dl>
 
     <dt>Dismissable</dt>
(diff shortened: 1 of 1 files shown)


joshueoconnor:
@steverep All looks good to me :-)

Issue #810, opened 2018-03-20
1.3.4 uses "meaning" 1.3.5 uses "purpose"


DavidMacDonald:
Is this a problem? 

I think we updated 1.3.4 to meaning, but didn't get to 1.3.5

Also, just need to be aware that 1.1.1 1st bullet says...
"Controls, Input: If non-text content is a control or accepts user input, then it has a **_name that describes its purpose._** " 

We may need to deal with that in the understanding... I can see a lot of confusion arising from 1.3.5 having "purpose" and 1.1.1 first bullet saying "purpose"

guyhickling:
Surely "meaning" is the wrong word to use in both SC1.3.4 and 1.3.5? The meaning of something is the information it conveys. Text can have a meaning. Images can have a meaning. But an input field is just a mechanism, a box for somebody to type into. It has no meaning of its own. A mechanism has a purpose, not a meaning.

So to use that word here will be confusing and could lead to future developers misunderstanding the SC. Some may think the meaning of the field is its label. They will say "I have given this field a &lt;label&gt; so the meaning can be programmatically determined, job done!

I would suggest changing SC1.3.4 back to "purpose", for clarity, not changing 1.3.5 in the other direction.

awkawk:
[Proposed Response]
We agree that 1.3.4 and 1.3.5 are intended to communicate the same concepts. Pull Request #856 (https://github.com/w3c/wcag21/pull/856) addresses this issue.

awkawk:
The WG decided on the above response, so we changed the text in the comment containing the proposed response to read "[Official WG Response]". Please confirm is you are satisfied with the response within 3 days. If we haven't heard a response by then we will regard the resolution as satisfactory.

PR #856, opened 2018-04-09
changes to make 1.3.4 align with 1.3.5 in use of purpose

awkawk:


awkawk:
@lseeman This is a pull request made in response to Issue 810 https://github.com/w3c/wcag21/issues/810

PR #864, opened 2018-04-13
Non text contrat mg

mbgower:
@alastc please review the changes proposed to date. Note that these are in additional to the changes I previously pulled on the main non-text-contrast branch (which are already incorporated into this branch), so those should be resolved first.


mbgower:
I made the pull against the master instead of against the branch.

PR #869, opened 2018-04-19
Renamed example 1 to remove "Visible Trigger" language

mbgower:
Added in additional text to explain what the images are showing.


DavidMacDonald:
I'm OK with #869

awkawk:
I'd like to have more generic examples, that is, not using the WCAG 2.1 GitHub repo as the source of the example. This example (#1) would also pass since it is implemented by the browser rather than the author.

awkawk:
I also think that we need to show an example of the intersection of tooltip hovering and screen magnification if we can.

PR #871, opened 2018-04-21
Deleted repeated sentence

DavidMacDonald:
This sentence was repeated "Regardless of the whether or not there is a visible indication of the hit area for the button, the focus indicator for the component must have sufficient contrast when the component is focused. "


DavidMacDonald:
Note: I would have preferred to edit this in the branch for the Understanding of non-text content, but it appears that that branch is behind the master.

PR #879, opened 2018-04-24
The current non-text contrast branch clean up

DavidMacDonald:


mbgower:
Hi @DavidMacDonald, Looking at it, this appears to be a version BEFORE my last set of edits. You have uncorrected many obvious typos, spelling usage issues, etc., that I corrected in my last pass. So I do not believe this pull request should take place. :grin: 
If it is still under consideration, I also want to know why you took out that material on the disabled components. I added that in specifically in response to issues that were opened about it -- including a very long discussion thread that had a lot of +1s, including from members of the working group. With that amount of discussion, I think an explanation about it is warranted.

Overall, i suspect this pull request should just be discarded. i am going to be reviewing and incorporating Bo's material, and if you want to do a pass _after_ that and see which of these changes you still want, that would be my preference.

alastc:
I'm wondering if we have branches taken off at different points, or perhaps David branched off master?

@DavidMacDonald could you check if the edits you thought you were making are showing in the diff?

It might be better to pull them into a new branch that comes off the non-text-contrast branch.

PR #884, opened 2018-04-27
Pointer cancellation mike

mbgower:
@detlevhfischer and @steverep . Please review this substantive change to the Understanding document. I believe I have addressed various comments given in Issues.


understanding/21/pointer-cancellation.html
```diff
@@ -9,28 +9,43 @@
...
+			<h3>Up-Event activation or completion</h3>
+			<p>The most accessible way to incorporate pointer cancellation is to make activation occur on the up-event.</p>
+     
+        <p>Up-event activation refers to the activation of a target when the pointer is released. In a touchscreen interaction, when the finger touches a target, the up-event activation only occurs when the finger is lifted while still being within the target boundary. Similarly in mouse interaction, the up-event occurs when the mouse button is released while the cursor is still within the boundary of the initial target set when the mouse button was pressed.</p>
+      <p>Authors can reduce the problem of users inadvertently triggering an action by using generic platform activation/click events that activate functionality on the up-event. For example, the "click" event in JavaScript triggers on release of the primary mouse button, and is an example of an implicit up-event. Despite its name, the onclick event handler is device-independent and also works for touch and keyboard interaction.</p>
```
patrickhlauke: tiny nitpick: might be better to talk about the "`click` event is device independent", rather than just the "onclick event handler" (as click is of course also device independent when attached via `addEventListener("click", ...)`

mbgower: @patrickhlauke, I am not enough of a developer to grasp the nuances, but it sounds like maybe we can use both?  
Does this make sense?
> Despite the names, both the <code>click</code> event and the <code>onclick</code> event handler are device-independent and also work for touch...

patrickhlauke: i don't really think you need to mention `onclick` per se at all? as it's simply the handler if you were doing old-school "event handlers as attributes in HTML" (a very discouraged practice) or doing something like `foo.onclick = ....` in JavaScript which is also quite oldschool compared to using `addEventListener`, since `onclick` would only allow a single function to be tied?

mbgower: So just:
> Despite the name, the <code>click</code> event is device-independent and also works for touch...

patrickhlauke: yup :)

understanding/21/pointer-cancellation.html
```diff
@@ -9,28 +9,43 @@
...
+         <p>The preference for up-events is implicit in the Success Criterion wording of the first bullet: <q>The down-event of the pointer is not used to execute any part of the function.</q> Authors meet the first bullet by using only the up-event.</p> 
+      <h3>Up-Event Abort or Undo</h3>
+       Where the interaction is equivalent to a simple "click", up-event activation has a built-in ability to cancel. There is a distinction between when someone touches a screen and when they remove their finger. Similarly, in mouse interaction, there is a difference between pressing and releasing the mouse button. When activation occurs only as the pointer is released, users have the opportunity to Abort (cancel) the activation. Users who have difficulty accurately using a mouse or touchscreen benefit greatly from this basic behaviour. They normally receive visual feedback when an item is pressed. If they discover they have selected the wrong item, they can cancel the action by moving their pointer or finger away from the target before releasing.</p>
```
alastc: Missing <p> tag?

PR #885, opened 2018-04-27
Non text contrast bo

mbgower:
Incorporated feedback from Bo and Lisa.


goodwitch:
Oh...I hope I didn't cause problems.  I just dropped in the Persona Quote for Non-Text Contrast.

mbgower:
Yeah, it looks to me like Glenda's modification happened after I made my branch of of the non-text-contrast branch. i don't see any conflicts since she touched a section i didn't. I think if her commit is backed out and then she just submits it again once my PR is handled, that should resolve, BUT I'm nothing like proficient in this. @michael-n-cooper can you have a look?

PR #890, opened 2018-05-01
Identify input purpose detlev

detlevhfischer:
Have reworked the text of "Understanding 1.3.4 Identify Input Purpose", including substantial rewording kindly provided by Patrick.  The Sufficient Techniques section has but one Technique so far. Not sure if other ways of adding programmatically determinable purpose are mature enough to be included at this  point.   I am uncertain whether the potential to furnish icons based on the autocomplete attribute should be played up here, or fit better into SC 1.3.5.
As to the Resources section, I would ask folks from the COGA TF to suggest sufficiently stable and relevant pointers.


alastc:
Hi Detlev,

Updates to understanding docs need to be merged into the 'working' branch, i.e. the id-input-purpose branch rather than master.

I don't think the target of the merge can be changed once created, so I created one that goes to the working branch in #891

detlevhfischer:
I did not intend to merge with master - sorry. Must read documentation - and hope it won‘t go wrong next time...

Sent from phone

PR #903, opened 2018-05-08
Proposed abstract for Rec

michael-n-cooper:
The abstract for the Recommendation should differ from the abstract for drafts, so needs updating. This pull request, although editorial, needs review from W3C management so is not ready to accept until after that.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag21/pull/903.html" title="Last updated on May 24, 2018, 8:25 PM GMT (6f8ba0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag21/903/8e8822f...6f8ba0c.html" title="Last updated on May 24, 2018, 8:25 PM GMT (6f8ba0c)">Diff</a>


brewerj:
Commenting on updated draft abstract with regard to potential differences in conformance between WCAG 2.1 and WCAG 2.0: 
- "Content that conforms to WCAG 2.1 also conforms to WCAG 2.0" -- this seems like a clear and important statement, which may work best as a standalone sentence.
- "..., and therefore, in the opinion of the Working Group, to policies that reference WCAG 2.0." -- this statement seems problematic in two respects; first, because different policies reference WCAG (2.0 and/or 2.1) in different ways, and it is unclear how the WG can directly claim that content conforming to WCAG 2.1 conforms to any given policy without knowing what that particular policy says, or what types of review a given policy went through. But also, even if so, it seems that the phrasing should be formulated differently, for instance "The WG intends that, for policies requiring conformance to WCAG 2.0, WCAG 2.1 can provide an alternate means of conformance." This approach might also address the first concern mentioned above.
- What still seems missing in this information though is a clear statement that W3C is not deprecating or superseding WCAG 2.0. Based on feedback to date, there already appear to be misunderstandings on this point, so this seems important to add. A candidate location to add this appears to be just before the next two sentences, e.g, "[add?] While W3C is neither deprecating nor superseding WCAG 2.0,..." then continuing with "...the W3C recommends that new and updated content use..." as is subsequently explained. IMO this clearly and appropriately states the implications of a W3C Recommendation. 
- The other possible location for this would be to add a simple statement of non-supersession for WCAG 2.0 in the status section of WCAG 2.1, e.g. "WCAG 2.0 is not superseded by the publication of WCAG 2.1." However, separating that relevant info from the abstract would potentially mean that it gets missed by some readers, so that positioning would probably be inadequate. And in addition, it would remove that information from the important context that is laid out in this part of the abstract, potentially leading to misunderstandings. Therefore I recommend positioning that clarification in the abstract as described above, so that the emphasis is on a clear statement of intent that WCAG 2.1 is recommended for new & updated content, and new and updated policies referencing WCAG.

Discussion welcome as needed; please let me know.

awkawk:
Added Judy's suggestions in this commit: https://github.com/w3c/wcag21/pull/903/commits/ece8492a33bc13bf11d8a458c407a7b03490965b

Also added one other change in the first paragraph:
Was:
Following these guidelines will make content more accessible to a wider range of people with disabilities, including blindness and low vision, deafness and hearing loss, limited movement, speech disabilities, photosensitivity, and combinations of these, and some accommodation for learning disabilities and cognitive limitations; but will not address every user need for people with these disabilities.

Changed to:
Following these guidelines will make content more accessible to a wider range of people with disabilities, including accommodations for blindness and low vision, deafness and hearing loss, limited movement, speech disabilities, photosensitivity, and combinations of these, and some accommodation for learning disabilities and cognitive limitations; but will not address every user need for people with these disabilities.

Could also be "including accommodations for people with blindness or low vision..."

mraccess77:
I'd prefer to not have the word "accommodation" or "accommodations" in the text.

awkawk:
@mraccess77 suggestions?

mraccess77:
Just remove the term "accommodations".

awkawk:
I thought that it sounded odd having: 
"Following these guidelines will make content more accessible to a wider range of people with disabilities, including blindness and low vision..."

It feels like we are intending to provide a list of examples of people with disabilities but instead list disabilities.

PR #922, opened 2018-05-15
Revised functionality as per meeting May 15th

DavidMacDonald:
Edited all requirements that imply that content needs to be fully tested in all orientation. SC is about orientation being locked, rather than ensuring that everything works in all orientations, which has compounding testing implications if it has to be tested after resizing SC is implemented, and reflow SC is implemented. This makes it a simple test. Turn scree orientation, does the page rotate?

### DIFF
--- a/understanding/21/orientation.html
+++ b/understanding/21/orientation.html
@@ -15,10 +15,10 @@
 		</h1>
 		<section id="intent">
 			<h2>Intent of this Success Criterion</h2>
-			<p>The intent of this Success Criterion is to ensure that content displays in the orientation (portrait or landscape) preferred by the user. Some websites and applications automatically set and lock the screen to a particular display orientation and expect
-			that users will respond by rotating their device to match. However, some users have their devices mounted
+			<p>The intent of this Success Criterion is to ensure that content displays in the orientation (portrait or landscape) preferred by the user. Some websites and applications automatically set and restrict the screen to a particular display orientation and expect
+			that users will respond by rotating their device to match, but this can create problems. Some users have their devices mounted
 			in a fixed orientation (e.g. on the arm of a power wheelchair). Therefore, websites and applications need to support both orientations
-			by not locking the orientation. Changes in content or functionality due to the size of display are not covered by this criteria which is focused on restrictions of orientation.</p>
+			by not restricting the orientation. Changes in content or functionality due to the size of display are not covered by this criteria which is focused on restrictions of orientation.</p>
 
       <p>Historically, devices tended to have a fixed-orientation display, and all content was created to match that display orientation. Today, most handhelds and many other devices (e.g., monitors) have a hardware-level ability to dynamically adjust default display orientation based on sensor information. The goal of this Success Criterion is that authors should never restrict content's orientation, thus ensuring that it always match the device display orientation.</p>
 
@@ -42,7 +42,7 @@
 			<h2>Examples</h2>
 			<ul>
 				<li><strong>Example 1: Online video site</strong><br />
-				Videos are not locked to a specific orientation.  A user can start a video with their device in portrait or in landscape orientation.</li>
+				A video is shown in either portrait or in landscape based on the user's chosen orientation.</li>
 				<li><strong>Example 2: Messaging website</strong><br />
 				A messaging website can display messages in both portrait and landscape orientations. </li>
 				<li><strong>Example 3: eReader web app</strong><br />


understanding/21/orientation.html
```diff
@@ -31,11 +31,11 @@ <h3>Benefits</h3>
 			<h2>Examples</h2>
 			<ul>
 				<li><strong>Example 1: Online video site</strong><br />
-				Videos are not locked to a specific orientation.  A user can start a video with their device in a portrait orientation, resulting in a smaller video being displayed and then rotate the device to landscape and have the video expand to fill the entire screen.  All controls and functionality are available in either orientation.</li>
+				Videos are not locked to a specific orientation.  A user can start a video with their device in a portrait orientation or in landscape.</li>
```
awkawk: I like the second sentence, think that we should just strike the "all controls...orientation" sentence.

understanding/21/orientation.html
```diff
@@ -31,11 +31,11 @@ <h3>Benefits</h3>
...
 				<li><strong>Example 2: Messaging website</strong><br />
 				A messaging website can display messages in both portrait and landscape orientations without a loss of content or functionality. </li>
 				<li><strong>Example 3: eReader web app</strong><br />
-				An eReader web app can display the contents of a book in both portrait and landscape mode without a loss of content or functionality</li>
+				An eReader web app can display the contents of a book in both portrait and landscape.</li>
```
awkawk: needs a word, either "mode" or "orientation".

Also, I think you missed the one above, see line 36.

understanding/21/orientation.html
```diff
@@ -15,10 +15,10 @@ <h1>
 		</h1>
 		<section id="intent">
 			<h2>Intent of this Success Criterion</h2>
-			<p>The intent of this Success Criterion is to ensure that content displays in the orientation (portrait or landscape) preferred by the user. Some websites and applications automatically set the screen to a particular display orientation and expect
+			<p>The intent of this Success Criterion is to ensure that the operation of the content has not been restricted by the author to a particular orientation (portrait or landscape) preferred by the user. Some websites and applications automatically set and lock the screen to a particular display orientation and expect
```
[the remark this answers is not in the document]
DavidMacDonald: So what we are saying in the Understanding is that the interpretation of the SC is that Operation=Functionality, and that every site that meets WCAG must demonstrate every SC in both orientations. 

I can't agree to that. It's way too much overhead and not what the SC was intended to do. THis was carefully negotiated and this is a requirements creep.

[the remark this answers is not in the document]
awkawk: I don't think that is what we are saying. We use the word operation in the SC, so shouldn't shy away from it. By saying that content can be operated in different orientations we are not saying that every bit of functionality in one orientation will be tested in all orientations.

understanding/21/orientation.html
```diff
@@ -15,10 +15,10 @@ <h1>
 		</h1>
 		<section id="intent">
 			<h2>Intent of this Success Criterion</h2>
-			<p>The intent of this Success Criterion is to ensure that content displays in the orientation (portrait or landscape) preferred by the user. Some websites and applications automatically set the screen to a particular display orientation and expect
+			<p>The intent of this Success Criterion is to ensure that content displays in the orientation (portrait or landscape) preferred by the user. Some websites and applications automatically set and lock the screen to a particular display orientation and expect
```
awkawk: I know that we discussed whether this is "locking orientation" or "not enabling awareness of changes in orientation". I think that the paragraph is clear without the "and lock" and the "not locking the" (below), as it says that both need to be supported. But I don't care that much, so whatever people decide...

understanding/21/orientation.html
```diff
@@ -31,11 +31,11 @@ <h3>Benefits</h3>
 			<h2>Examples</h2>
 			<ul>
 				<li><strong>Example 1: Online video site</strong><br />
-				Videos are not locked to a specific orientation.  A user can start a video with their device in a portrait orientation, resulting in a smaller video being displayed and then rotate the device to landscape and have the video expand to fill the entire screen.  All controls and functionality are available in either orientation.</li>
+				Videos are not locked to a specific orientation.  A user can start a video with their device in portrait or in landscape orientation.</li>
```
alastc: Comment from MichaelG: "A video is shown in either portrait or in landscape based on the user's chosen orientation."

understanding/21/orientation.html
alastc: From meeting: Happy with 'set and lock', would like to add this on line 19: 
"...their device to match, but this can create problems. Some users..."

PR #924, opened 2018-05-16
Non text contrast ac

alastc:
Added more (draft) examples, updated the structure. Mostly editorial and obvious graphics updates.
Two key points though:
1. Removed the bit about relative luminance in the 1.4.1 section. I don't think it is an agreed interpretation that each state has to differ from the other states.
2. Added a possibly controversial example (4) in the inputs table, similar to an example from the Material Design library.


understanding/21/non-text-contrast.html
```diff
@@ -40,46 +59,49 @@ <h1>Understanding Non-text Contrast</h1>
...
+					<h4>Relation to <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html">1.4.1 Use of Color</a></h4>
+					<p>Designers will often use color to indicate the states of UI components. So how does the guidance on Use of Color relate to Non-text Contrast?</p>
+					<p> The criteria for Use of Color states <q>Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.</q> Although every object on a screen is arguably created through use of color only (shapes are created by altering an area of pixels from one color to another), the context of 1.4.1 addresses changing <strong>only the color</strong> of an object (or text) without otherwise altering the object's form. If an author adds a border, outline, or shadow to a shape to indicate a change in state then it is not relying solely on use of color.</p>
```
awkawk: The <a href=link>Use of Color</a> success criteria states ...

understanding/21/non-text-contrast.html
```diff
@@ -210,58 +241,109 @@ <h4>Gradients</h4>
...
+							</figcaption>
+						</figure>
+						<p>There are many possible solutions to ensuring contrast, the example shows the use of borders. Other techniques are to use darker colours for the circle backgrounds, or to add text labels &amp; values for each item.</p>
+					</section>
+				</section><!-- end graphic-object-examples -->
+
+			</section><!-- end graphical-objects -->
+
+			<section id="testing-principles">
+					<h3>Testing Principles</h3>
```
awkawk: I'm not entirely sold on this section. I can live with it, but think that this understanding document is already long and this moves into the techniques territory, so I would like to figure out how to integrate it into the structure better or leave it for techniques, sooner or later.

alastc: If someone can think of a better way of doing that, happy to adjust. The main things is that the SC covers two quite different elements on the page, each of which have multiple ways to meet the SC, so no structure of and/or for the techniques will address this.

awkawk:
@alastc once your images are updated then you can go ahead and merge.

alastc:
Thanks, merging to the main branch, I'll delete this branch so any further updates can happen in the main branch.

PR #944, opened 2018-05-28
Image compression (routine optimization)

j9t:
Shaved some KB off a few images. Part of a routine run over local repos.


awkawk:
thanks!

Issue #914, opened 2018-05-11
SC 1.4.11 boundaries and states of user interface components


goetsu:
Hello,

I'm a bit confuse because after looking at this SC and at websites in the implementation list 
https://www.w3.org/WAI/WCAG21/implementation-report/implementation_list?unit=non-text-contrast&category=sc either I don't understand the boundaries and states case correctly either some of those website aren't compliant.

Maybe because a definition of "visual boundaries of the component" is missing in the understanding

for example with the interpretation I made of visual boundaries : 
https://www.funka.com/en/ : 
* contrast of the search field against grey background of the header isn't compliant
* contrast of the bottom border of most of the news isn't compliant with white background

https://www.statefarm.com/ : 
* contrast of active state of the menu button when the menu is open again the red background isn't compliant
* contrast of zip code input border (light grey) against white isn't compliant

http://adrianroselli.com/2018/02/github-contributions-chart.html
* contrast of border of navigation item (dark grey) against black isn't compliant

https://knowbility.org/
* contrast of the burger icon (mobile version) background against white background isn't compliant

http://www.lflegal.com/
* contrast of navigation item with grey border against light grey background below them isn't compliant

https://www.microassist.com/project/viral-load-changes-hiv/
* contrast of border of navigation item inside menu that appear on hover isn't compliant
* contrast of border of the search field against white background isn't compliant (neither yellow magnifying glass icon)

Also the understanding document state "the visual boundaries of the component must have sufficient contrast with the adjacent background" I think it must be "the visual boundaries of the component must have sufficient contrast with the external adjacent background on all sides of the component"
Otherwise it mean that if you have a light grey background input with a grey border on a white background the light grey background of the input itself must also have a contrast ratio of 3 at least against grey border

swickr:
For reference: there is AG WG discussion in a [mail list thread](https://lists.w3.org/Archives/Public/w3c-wai-gl/2018AprJun/0542.html)

awkawk:
[Proposed Response]:

@goetsu 
For context, please take a look at the updated SC text. The Working Group has not changed its interpretation of this SC but in response to comments and questions has updated the text to better reflect the WG's intent.

SC text:
https://w3c.github.io/wcag21/guidelines/#non-text-contrast

> https://www.funka.com/en/ :
> contrast of the search field against grey background of the header isn't compliant
contrast of the bottom border of most of the news isn't compliant with white background

[...]

PR #945, opened 2018-05-29
change per 5/29/18 call

awkawk:


swickr:
Additional background in #914

